PR #27, opened 2023-10-03
feat: deployment_config should define pipelines as a dict instead of list

Barabazs:
With this PR pipelines should now be defined as a dictionary of pipelines under the environment top-level key.
This has the implication that it's not possible to define pipelines a string in a deploy.yml file.


Barabazs:
Canceling due to unacceptable trade-off.

PR #37, opened 2023-10-05
Derive pipeline_name from loaded config

lionellloh:
## Run a pipeline 

```
aineko run -c ./conf/pipeline.yml
``` 

## Before: (says None pipeline) 
```
All datasets created.
2023-10-05 08:48:11,403	INFO worker.py:1633 -- Started a local Ray instance. View the dashboard at 127.0.0.1:8266
Running sequence node on None pipeline: inputs=None, outputs=['test_sequence', 'logging']
Running sum node on None  pipeline: inputs=['test_sequence'], outputs=['test_sum', 'logging']
aineko.core.node_manager.NodeManager
``` 

## After:  (gives pipeline name) 
```
All datasets created.
2023-10-05 08:48:52,254	INFO worker.py:1633 -- Started a local Ray instance. View the dashboard at 127.0.0.1:8266Running sequence node on test-aineko-pipeline pipeline: inputs=None, outputs=['test_sequence', 'logging']
Running sum node on test-aineko-pipeline pipeline: inputs=['test_sequence'], outputs=['test_sum', 'logging']
```


craycrayfish:
Looks good, thanks! Can we just change the phrasing to
```
Running node `sequence` on pipeline `test-aineko-pipeline`: ...

craycrayfish (approved):

Barabazs (commented):
I'm not going to block, but I disagree about the location where we start inferring the name of the pipeline. Imo we should do this in the config_loader, because that allows us to load and parse a config without needing to run the actual pipeline.

lionellloh:
> I'm not going to block, but I disagree about the location where we start inferring the name of the pipeline. Imo we should do this in the config_loader, because that allows us to load and parse a config without needing to run the actual pipeline.

I agree with the sentiment. But ConfigLoader is kind of the module that does the inferring since it loads the config. I could modify the return of `.load_config()` to return a tuple[pipeline_name: str, pipeline: dict] but pipeline_name is already in that dict. We are just setting the object state `self.pipeline_name` in the runner with the returned dict in this case.

PR #39, opened 2023-10-06
Fix/standardize cli config param

lionellloh:
## Before: 
```
aineko run -c ../example_pipelines/my_awesome_pipeline/conf/pipeline.yml
```

## After: 
```
aineko run ../example_pipelines/my_awesome_pipeline/conf/pipeline.yml
```

## Context:

No point passing in an option if we always need it. This also has the added benefit of standardizing CLI usage with the `visualize` command


Barabazs:
Can you always branch of develop please? 
And if you need code change from branch A, then imo this PR should merge into branch A instead of develop.

Also, please try to use conventional commits.

Barabazs (approved):

lionellloh:
> Can you always branch of develop please?
> And if you need code change from branch A, then imo this PR should merge into branch A instead of develop.

I agree! My bad.  I use hg most of the time and tbh one thing I haven't quite figure out is how best to stack changes. (If a PR is in-review, how can I continue making changes on top of that efficiently? I guess branching off that PR is one way.) 

> Also, please try to use conventional commits.

Sure, do you have a guide you can point me to? Do you mean using qualifiers like `(chore)`, `(lint)` etc? My past experiences with OS is that the commit message just have to be descriptive.

craycrayfish:
> Sure, do you have a guide you can point me to? Do you mean using qualifiers like `(chore)`, `(lint)` etc? My past experiences with OS is that the commit message just have to be descriptive.

Refer to this for conventional commits: https://www.conventionalcommits.org/en/v1.0.0/

craycrayfish (approved):

Barabazs:
> I agree! My bad. I use hg most of the time 

I forgot you guys use Mercurial.

> (If a PR is in-review, how can I continue making changes on top of that efficiently? I guess branching off that PR is one way.)

Yes, that's what I usually do. This can get "messy" when the initial branch is modified because you'll need to update the new branch too. In that case I mostly end up rebasing or merging.

craycrayfish (approved):

PR #43, opened 2023-10-18
feat: added prefixing capabilities

craycrayfish:
This PR adds the ability to namespace kafka topics created by the datasets.
Datasets defined will have a prefix containing the pipeline name added, and an optional universal prefix that can be defined by the user. Topic names will be:
```
<universal_prefix>.<pipeline_name>.<dataset_name>
```

### DIFF
--- a/aineko/core/dataset.py
+++ b/aineko/core/dataset.py
@@ -89,7 +89,7 @@ class DatasetConsumer:
                 "group.id"
             ] = f"{prefix}.{pipeline_name}.{node_name}"
             self.consumer = Consumer(consumer_config)
-            self.consumer.subscribe([f"{prefix}.topic_name"])
+            self.consumer.subscribe([f"{prefix}.{topic_name}"])
 
         else:
             consumer_config["group.id"] = f"{pipeline_name}.{node_name}"
--- a/aineko/core/dataset.py
+++ b/aineko/core/dataset.py
@@ -96,6 +96,8 @@ class DatasetConsumer:
             self.consumer = Consumer(consumer_config)
             self.consumer.subscribe([topic_name])
 
+        self.topic_name = topic_name
+
     @staticmethod
     def _validate_message(
         message: Optional[Message] = None,
--- a/aineko/core/dataset.py
+++ b/aineko/core/dataset.py
@@ -32,9 +32,11 @@ class DatasetConsumer:
     """Wrapper class for Kafka consumer object.
 
     DatasetConsumer objects are designed to consume messages from a single
-    dataset and can consume messages in two different ways:
-        "next": read the next unconsumed message in the queue
-        "last": read the latest message in the queue
+    dataset and will consume the next unconsumed message in the queue.
+    Prefixes will automatically be added to the dataset name. For example,
+    a dataset named "dataset_1" in a pipeline named "pipeline_1" with a prefix
+    of "test" will subscribe to a topic named "test.pipeline_1.dataset_1" if
+    `has_pipeline_prefix` is set to `True`, and "test.dataset_1" otherwise.
 
     Args:
         dataset_name: name of the dataset
@@ -42,6 +44,8 @@ class DatasetConsumer:
         pipeline_name: name of the pipeline
         dataset_config: dataset config
         broker: broker to connect to (ip and port: "54.88.142.21:9092")
+        prefix: prefix for topic name (<prefix>.<dataset_name>)
+        has_pipeline_prefix: whether the dataset name has pipeline name prefix
 
     Attributes:
         consumer: Kafka consumer object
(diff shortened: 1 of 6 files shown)


tests/integration/test_runner_with_kafka.py
```diff
@@ -58,6 +58,8 @@ def _execute(self, params: Optional[dict] = None) -> None:
         """Read message"""
         msg = self.consumers["messages"].consume()
         if time.time() - self.last_updated > self.timeout:
+            print(f"Received messages: {self.received}")
+            print(self.consumers["messages"].topic_name)
```
Barabazs, line 62: I don't think this will ever be visible in the console when running with pytest.

aineko/core/node.py
```diff
@@ -111,28 +113,39 @@ def setup_datasets(
             pipeline: name of the pipeline
             inputs: list of dataset names for the inputs to the node
             outputs: list of dataset names for the outputs of the node
+            prefix: prefix for topic name (<prefix>.<dataset_name>)
```
Barabazs:
```suggestion
            prefix: prefix for topic name (`<prefix>.<dataset_name>`)
```
For some reason VS Code doesn't render `(<prefix>.<dataset_name>)` correctly.
image

aineko/core/dataset.py
```diff
@@ -56,33 +60,43 @@ def __init__(
...
+        topic_name = dataset_name
+        if has_pipeline_prefix:
+            topic_name = f"{pipeline_name}.{dataset_name}"
+
+        if self.prefix:
+            consumer_config[
+                "group.id"
+            ] = f"{prefix}.{pipeline_name}.{node_name}"
+            self.consumer = Consumer(consumer_config)
+            self.consumer.subscribe([f"{prefix}.topic_name"])
```
Barabazs: Should it not be:
```suggestion
            self.consumer.subscribe([f"{prefix}.{topic_name}"])
```

aineko/core/dataset.py
```diff
@@ -56,33 +60,43 @@ def __init__(
...
         # Override default config with dataset specific config
         for param, value in dataset_config.get("params", {}).items():
             if param in self.kafka_config.get("CONSUMER_OVERRIDABLES"):
                 consumer_config[param] = dataset_config["params"][value]
 
-        # Add consumer group id based on node name
-        consumer_config["group.id"] = f"{self.pipeline_name}.{node_name}"
+        topic_name = dataset_name
+        if has_pipeline_prefix:
+            topic_name = f"{pipeline_name}.{dataset_name}"
```
Barabazs, line 90: Why are you using topic_name here instead of dataset_name (or a variant)?

Barabazs (commented):
I left a few notes, but overall I'm confused by the explanation (in the docstrings) and use of prefix and has_pipeline_prefix.

tests/integration/test_runner_with_kafka.py
craycrayfish: It was visible when I was testing it earlier

aineko/core/dataset.py
craycrayfish: Because there is now a separation of the concept of topic name and dataset name. A dataset name is the name within the namespace of the pipeline, so will be the short version that the user uses in pipeline config and node code. The topic name is the actual name used to call the kafka topics, and needs to be in the project namespace, so will contain the pipeline name as a prefix. 
E.g.
a dataset is called `count`
the associated topic is really named `my_pipeline.count`

craycrayfish:
> I left a few notes, but overall I'm confused by the explanation (in the docstrings) and use of prefix and has_pipeline_prefix.

I tried to make it clearer. Let me know if you're still confused and we can hop on a call to figure out how to make it understandable.

aineko/core/dataset.py
```diff
@@ -56,33 +65,43 @@ def __init__(
         node_name: str,
         pipeline_name: str,
         dataset_config: Dict[str, Any],
-        broker: Optional[str] = None,
+        bootstrap_servers: Optional[str] = None,
+        prefix: Optional[str] = None,
+        has_pipeline_prefix: bool = False,
```
mdigi14, line 70: I don't think we need `has_pipeline_prefix`, unless we expect the user to set `prefix=""` (which is probably a case we should handle either way)

If `prefix` defaults to `None` then `if has_pipeline_prefix:` can be replaced by `if prefix:`

mdigi14: Or to follow the Google Python Style Guide, we can do 

`if prefix is not None:`

mdigi14: And then maybe raise a `ValueError` for empty strings?

mdigi14 (changes requested):
Left some comments related to the `has_pipeline_prefix` variable

aineko/core/dataset.py
craycrayfish: I'm intentionally creating 2 sets of prefixes: `prefix` and `pipeline_prefix`. `prefix` can be any string defined by the user upon pipeline creation. `pipeline_prefix` adds the pipeline name as a prefix to datasets.

`pipeline_prefix` is used for all user-defined datasets but not the `logging` topic. It adds the pipeline name as a prefix so the topic is called `my_pipeline.my_dataset`.

`prefix` is not currently used, but will be useful for adding a universal prefix to ALL datasets associated to this aineko project, so that one can namespace the kafka topics associated with this project. This will be required in a multi-tenant kafka setup, but not for a single-tenant setup. It can add any string so the topic will be called `my_prefix.my_pipeline.my_dataset`.

craycrayfish: There's 2 levels of namespacing going on.
 
- Project-level namespace: defined by the `prefix`.
- Pipeline-level namespace: on by default, defined by `has_pipeline_prefix`

mdigi14: Oh I see, my bad. I think it might be better to call the variable `use_pipeline_prefix` but I'll approve.

mdigi14 (approved):

PR #45, opened 2023-10-20
feat: implement method for testing the node state, input & output at each iteration

Barabazs:
This PR implements a new method for unit testing nodes. The main difference with the current `run_test` method is that this method yields the input, output and node instance at each iteration of the loop.  This allows us to:
* test intermediate outputs instead of the aggregated output at the end of the loop
* access the internal state of the node at each iteration
* access the node methods at each iteration

### DIFF
--- a/aineko/core/node.py
+++ b/aineko/core/node.py
@@ -17,7 +17,7 @@ that users should override with their own implementation.
 import time
 import traceback
 from abc import ABC, abstractmethod
-from typing import Dict, List, Optional
+from typing import Dict, Generator, List, Optional, Tuple
 
 import ray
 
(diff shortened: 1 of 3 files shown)


ricrosales:
This is cool! Correct me if I'm wrong, but it seems like this does not break the current testing framework right?

Barabazs:
> This is cool! Correct me if I'm wrong, but it seems like this does not break the current testing framework right?


Correct. It's a totally separate method and both testing methods can be used.

craycrayfish:
Should we add an extra argument in `run_test` that enables this mode rather than use a different method altogether?

Barabazs:
> Should we add an extra argument in `run_test` that enables this mode rather than use a different method altogether?

I actually started with this and then I realised it got messy to combine yielding and returning values. We could however rename the current  run_test to run_test_classic. And create a new run_test function that calls the appropriate function based on a boolean.

```
def run_test(iterative_yield: False):
    if iterative_yield:
        run_test_yield(...)
    else:
        run_test_classic(...)
```

craycrayfish (approved):
Nice!

PR #46, opened 2023-10-23
refactor: replace Schema with Pydantic model

Barabazs:
This PR replaces the Schema definition in config_loader with a Pydantic model.
The docstrings were stale and have also been updated to reflect the fact that we only load one pipeline config and we don't use projects anymore.


craycrayfish (approved):
Very nice!

PR #70, opened 2023-11-07
Refactor/move to click

craycrayfish:
This PR migrates the CLI command to the python `click` framework. This provides the following benefits:
- More compact code
- Automated generation of CLI documentation with `mkdocs`
- More intuitive code generation
- Tests for CLI commands 

CI config has been updated to run tests marked as `integration` separately, as those have a dependency on having docker available.

PR #71, opened 2023-11-07
feat: revamp documentation with mkdocs

craycrayfish:
This PR migrates documentation from Gitbook to MD files used by [`mkdocs`](https://www.mkdocs.org/) to generate a static documentation site. We use [`Material for MKDocs`](https://squidfunk.github.io/mkdocs-material/) due to the added features. 

Files for the documentation site is contained in the `docs-site` branch.
Upon merge into the `develop` branch, our CI will build and push a new change for the `dev` version.
Upon tagging with a valid semver version such as `1.2.3`, our CI will build and push a new change for the minor version (e.g. `1.2.`).
This will automatically be deployed by AWS Amplify.

An initial publishing of the site can be found on: https://docs.aineko.dev


mkdocs.yml
Barabazs, line 1: Can you add versioning please?
https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/

.circleci/config.yml
```diff
@@ -24,6 +24,12 @@ workflows:
           python_version: << pipeline.parameters.python_version >>
           markers: integration
 
+  publish-docs:
+    when:
+      equal: [<< pipeline.git.branch >>, develop]
+    jobs:
+      - build-and-push-docs
+
```
Barabazs: I don't think we should always publish docs when merging to develop. We need to make a distinction between general changes that don't apply to a specific version (like correction, clarifications, formatting,...) and updates for unreleased code.

README.md
Barabazs, line 1: I don't really like to repeat the same information in many different places. This introduces the risk of updating it in one place and forget about other places. Can we just link to the docs/quickstart.md ?

aineko/cli/cli.py
Barabazs, line 1: This is not the same CLI code that was approved in PR #70 and merged to develop

Barabazs (changes requested):
Thank you for this amazing work (and in record time!)

I left a few comments that imo should be addressed before merging.

README.md
craycrayfish: I don't like it too, but I could not figure out a way to do that. mkdocs does not allow both an index.md and readme.md to co-exist.

Barabazs: We could just link to the relevant parts.

.circleci/config.yml
craycrayfish: Then what would be the ruleset we use to determine if we should publish docs?

Barabazs: On second thought, If we use versioning we can always publish. But it's not going to show up on the website, since it's not tied to a version.

README.md
giddyphysicist: would it be overkill to add a pre-commit-hook to copy the docs/quickstart.md file into the top level repo README file?

Barabazs (approved):
I updated the PR title to be more descriptive, but otherwise this is ready to be merged imo. 

Try to delete all the commit messages when you squash merge this please.

PR #112, opened 2023-12-20
feat: add error handling for invalid offset

craycrayfish:
## Summary

Add error handling where we skip attempting to assign partition offsets if they are invalid.

## Motivation

Issues were observed where deployments could not assign partitions when the `get_watermark_offsets` method was returning invalid offset values.

## Author Checklist

- [x] Status checks pass.

## Reviewer Checklist

### DIFF
--- a/aineko/core/dataset.py
+++ b/aineko/core/dataset.py
@@ -21,7 +21,6 @@ e.g. message:
 import datetime
 import json
 import logging
-import time
 from typing import Any, Dict, Literal, Optional
 
 from confluent_kafka import (  # type: ignore
--- a/aineko/core/dataset.py
+++ b/aineko/core/dataset.py
@@ -21,9 +21,11 @@ e.g. message:
 import datetime
 import json
 import logging
+import time
 from typing import Any, Dict, Literal, Optional
 
 from confluent_kafka import (  # type: ignore
+    OFFSET_INVALID,
     Consumer,
     KafkaError,
     Message,
@@ -100,13 +102,13 @@ class DatasetConsumer:
             topic_name = f"{pipeline_name}.{dataset_name}"
 
         if self.prefix:
-            consumer_config[
-                "group.id"
-            ] = f"{prefix}.{pipeline_name}.{node_name}"
+            self.name = f"{prefix}.{pipeline_name}.{node_name}"
+            consumer_config["group.id"] = self.name
             self.consumer = Consumer(consumer_config)
             self.consumer.subscribe([f"{prefix}.{topic_name}"])
 
         else:
+            self.name = f"{pipeline_name}.{node_name}"
             consumer_config["group.id"] = f"{pipeline_name}.{node_name}"
             self.consumer = Consumer(consumer_config)
             self.consumer.subscribe([topic_name])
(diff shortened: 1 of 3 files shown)


Barabazs (commented):
Can you add test cases that validate this fix?

aineko/core/dataset.py
```diff
@@ -153,12 +154,19 @@ def _update_offset_to_latest(self) -> None:
...
         for partition in partitions:
-            partition.offset = (
-                self.consumer.get_watermark_offsets(
-                    partition, cached=self.cached
-                )[1]
-                - 1
-            )
+            high_offset = self.consumer.get_watermark_offsets(
+                partition, cached=self.cached
+            )[1]
```
giddyphysicist, line 159: is there any chance this line could fail? for example, if get_watermark_offsets didn't return a tuple, then the [1] would error. if the method is strict in what it can return, then it should be fine

mdigi14: We should handle this upstream in `_update_offset_to_latest`. If this function fails because `get_watermark_offsets()` returns a `KafkaException` then I think we could either gracefully fail by using the same offset as the last call or raise a fatal error.

craycrayfish:
> Can you add test cases that validate this fix?

Added a test to check that these methods work when calling on an empty dataset, which was an edge case we suspected was causing this issue. Admittedly it doesn't cover the full spectrum of possible kafka scenarios.

aineko/core/dataset.py
craycrayfish: Added an error case where the message defaults to the next message if offset cannot be updated.

tests/integration/test_kafka_edge_cases.py
```diff
@@ -0,0 +1,48 @@
...
+    except ray.exceptions.RayActorError:
+        consumer = DatasetConsumer(
+            dataset_name="test_result",
+            node_name="consumer",
+            pipeline_name="integration_test_kafka_edge_cases",
+            dataset_config={},
+            has_pipeline_prefix=True,
+        )
+        count_messages = consumer.consume_all(end_message="END")
+        assert count_messages[0]["message"] == "OK"
```
Barabazs: At first glance I was confused by what you're doing here. I don't think it's very common to use a try-except like that. I would suggest refactoring it so it's easier to understand to everyone.

```suggestion
    try:
        runner.run()
    except ray.exceptions.RayActorError:
        # This is expected because we activated the poison pill
        pass

    consumer = DatasetConsumer(
        dataset_name="test_result",
        node_name="consumer",
        pipeline_name="integration_test_kafka_edge_cases",
        dataset_config={},
        has_pipeline_prefix=True,
    )
    count_messages = consumer.consume_all(end_message="END")
    assert count_messages[0]["message"] == "OK"
```

Barabazs: Also fwiw this test passes on the current develop, so without your fix. I assume this is because the dataset wasn't purged locally since I first ran the test on this branch and then on develop. If that's the case, our tests would need to do some clean-up before and after running. What do you think?

Barabazs (approved):
I left a non-blocking suggestion.

tests/integration/test_kafka_edge_cases.py
mdigi14: I agree with Erwin on not putting logic inside an `except` block. It should only be used for error handling.

mdigi14 (approved):

PR #120, opened 2023-12-28
feat: implement log_to_dataset to read python native logs and produce them to the logging dataset - BLD-1212

Barabazs:
## Summary

This PR implements a new feature to read logs that are produced outside of a node using the native python logging module.

## Motivation

Catching logs from outside of the nodes can be very useful for debugging or monitoring.

## Author Checklist

- [x] Status checks pass.

## Reviewer Checklist

### DIFF
--- a/docs/developer_guide/node_implementation.md
+++ b/docs/developer_guide/node_implementation.md
@@ -169,13 +169,59 @@ The timeout argument in these methods signify the duration in which the method h
 
 
 ### Logging
-
-Node classes inherit a method named `self.log` that allows users to log messages. You can set the appropriate level from: `info`, `debug`, `warning`, `error`, an `critical`. You can log from inside of the `_pre_loop_hook` method, the `_execute` method, or any other method you add to your node.
+#### From within a node
+Node classes inherit a method named `self.log` that allows users to produce messages to the `logging` dataset. This is useful for debugging and monitoring purposes. You can specify the level of the log message, by specifying the `level` argument. By default, the level is set to `info`. Other valid levels are `debug`, `warning`, `error`, and `critical`. You can log from any method that inherits from `AbstractNode`.
 
 :   
     ```python
     self.log(f"Produced {self.cur_integer}", level="info")
     ```
+
+#### From outside a node
+Since `self.log` is inherited from `AbstractNode`, you can't use it outside of a node. However, you can still log to the `logging` dataset by using the native Python `logging` module and specifying the `logging_namespace` and setting `log_to_dataset` to `true` in the node or pipeline configuration as shown below. This configuration will read all log entries from the "foo" namespace and produce them to the `logging` dataset.
+
+:   
+
+    ```yaml title="conf/pipeline.yml" linenums="7"
+    nodes:
+      example_node:
+        class: your_package.your_module.YourNode
+        log_to_dataset: true
+        logging_namespace: foo
+    ```
+
+    ```python title="your_package/your_utils.py" linenums="7"
+    import logging
+
+    logger = logging.getLogger("foo")
+    
+    def some_function():
+        logger.info("Hello world!")
+    ```
+ 
+    ```python title="your_package/your_module.py"
+    from your_package.your_utils import some_function
+    from aineko.core.node import AbstractNode
+
+    class YourNode(AbstractNode):
+        def _execute(self, params=None):
+            some_function()
+    ```
+??? note
+    * In the preceding example, the `logging_namespace` is set to `foo`. This means that only log entries with the namespace "foo" will be read and produced to the `logging` dataset. If you want to read and produce all log entries, set `logging_namespace` to `None`.
+    * Since nodes are isolated processes, the preceding example will only read logs that are written when invoked (either directly or indirectly) by the node. Specifying the logging parameters at the node level allows for very fine-grained control over what log messages are read and produced. However, the same parameters can be specified at the pipeline level. This will read all log entries from the specified namespace and produce them to the `logging` dataset.
+
+    :   
+        ```yaml title="conf/pipeline.yml" linenums="7"
+        nodes:
+          example_node:
+            class: your_package.your_module.YourNode
+          second_node:
+            class: your_package.your_module.YourSecondNode        
+           
+        log_to_dataset: true
+        logging_namespace: foo
+        ```
 
 
 ### PoisonPill
--- a/docs/developer_guide/node_implementation.md
+++ b/docs/developer_guide/node_implementation.md
@@ -222,6 +222,7 @@ Since `self.log` is inherited from `AbstractNode`, you can't use it outside of a
         log_to_dataset: true
         logging_namespace: foo
         ```
+    * Logging parameters at the node level have precedence over logging parameters at the pipeline level. For example, if `log_to_dataset` is set to `true` at the pipeline level, but `log_to_dataset` is set to `false` for node `xyz`, then node `xyz` will nog read and produce log entries. However, all other nodes will read and produce log entries.
 
 
 ### PoisonPill
--- a/aineko/core/node.py
+++ b/aineko/core/node.py
@@ -87,7 +87,7 @@ class AbstractNode(ABC):
         poison_pill: Optional[ray.actor.ActorHandle] = None,
         test: bool = False,
         logging_namespace: Optional[str] = None,
-        log_to_dataset: bool = False,
+        log_to_dataset: Optional[bool] = None,
     ) -> None:
         """Initialize the node."""
         self.name = node_name or self.__class__.__name__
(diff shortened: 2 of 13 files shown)


mdigi14:
This is a really good feature, but it doesn't feel logical to configure this inside of node config because this is something that would typically be used outside of a node. Do you think there is a good reason not to just have this be a pipeline-level config?

I guess this allows the user to user `print()` inside of a node, but I'm not sure if we want to have both `print()` and `self.log()` available inside a node.

Barabazs:
> This is a really good feature, but it doesn't feel logical to configure this inside of node config because this is something that would typically be used outside of a node. Do you think there is a good reason not to just have this be a pipeline-level config?

This allows the user to be very granular in which nodes capture logs or not. But the user can also configure it at the pipeline-level.

> I guess this allows the user to user `print()` inside of a node, but I'm not sure if we want to have both `print()` and `self.log()` available inside a node.

No, `print` will not be captured.

docs/developer_guide/node_implementation.md
```diff
@@ -169,14 +169,61 @@ The timeout argument in these methods signify the duration in which the method h
...
+      example_node:
+        class: your_package.your_module.YourNode
+        log_to_dataset: true
+        logging_namespace: foo
+    ```
+
+    ```python title="your_package/your_utils.py" linenums="7"
+    import logging
+
+    logger = logging.getLogger("foo")
```
craycrayfish, line 196: I think it's confusing that people have to create a logger namespace in a node before they can use it anywhere. Can we do something like creating a default namespace called `kafka_logging` or `logging` and create `logger = logging.getLogger("foo")` in `aineko/__init__.py` so users can do `from aineko import logging` without having to create it first. 

This doesn't stop them from creating their own loggers.

craycrayfish: maybe we can use the NodeManager or one of the internal node's self.log method for this. Or implement an independent logger that doesn't require a self.log

craycrayfish: If we make it dependent on a `self.log` method call, this will be unable to capture logs that occur before a pipeline goes up and after a pipeline goes down. I think having an independent dataset producer object that writes to the logging topic might be better

craycrayfish:
Pretty elegant implementation, left some suggestions on easier UX

docs/developer_guide/node_implementation.md
Barabazs: > I think it's confusing that people have to create a logger namespace in a node before they can use it anywhere. 

Afaik it's a common/best practice to initialise a logger with `logger = logging.getLogger(__name__) `  at the top of each module. This allows for hierarchical loggers as described in [Logger Objects](https://docs.python.org/3/library/logging.html#logger-objects).

```
2024-01-02 09:43:48.000 - your_package.your_module_1 - INFO - Hello from module 1!
2024-01-02 09:43:49.000 - your_package.your_module_2 - INFO - Hello from module 2!
2024-01-02 09:43:50.000 - your_package.your_module_1 - ERROR - Hello from module 1!
```

> Can we do something like creating a default namespace called `kafka_logging` or `logging` and create `logger = logging.getLogger("foo")` in `aineko/__init__.py` so users can do `from aineko import logging` without having to create it first.

I think it would make the feature less useful since users would never know where the log came from. They would all originate from "foo".

```
2024-01-02 09:43:48.458 - foo - INFO - Hello, World!
```


EDIT: Maybe I should update the docs to use `__name__` instead of `"foo"`. What do you think?

Barabazs: > maybe we can use the NodeManager or one of the internal node's self.log method for this. Or implement an independent logger that doesn't require a self.log

> If we make it dependent on a `self.log` method call, this will be unable to capture logs that occur before a pipeline goes up and after a pipeline goes down. I think having an independent dataset producer object that writes to the logging topic might be better

How do you envision this independent object? 

The complexity of logging before/during a pipeline starting lies in the distributed nature of the processes and the creation of the topics. Everything that happens before a node starts isn't visible to the node, so it can't capture and produce. But maybe we can store these logs temporarily until the NodeManager is running and then produce them to the logging topic. It's not a trivial feature, so maybe we should spec that out separately.

mdigi14:
> > This is a really good feature, but it doesn't feel logical to configure this inside of node config because this is something that would typically be used outside of a node. Do you think there is a good reason not to just have this be a pipeline-level config?
> 
> This allows the user to be very granular in which nodes capture logs or not. But the user can also configure it at the pipeline-level.
> 
> > I guess this allows the user to user `print()` inside of a node, but I'm not sure if we want to have both `print()` and `self.log()` available inside a node.
> 
> No, `print` will not be captured.

@Barabazs Isn't it confusing if the user can use both `self.log()` and `logger = logging.getLogger("foo")` inside of a node? what is the use case where a user would use both?

Barabazs:
> > > This is a really good feature, but it doesn't feel logical to configure this inside of node config because this is something that would typically be used outside of a node. Do you think there is a good reason not to just have this be a pipeline-level config?
> > 
> > 
> > This allows the user to be very granular in which nodes capture logs or not. But the user can also configure it at the pipeline-level.
> > > I guess this allows the user to user `print()` inside of a node, but I'm not sure if we want to have both `print()` and `self.log()` available inside a node.
> > 
> > 
> > No, `print` will not be captured.
> 
> @Barabazs Isn't it confusing if the user can use both `self.log()` and `logger = logging.getLogger("foo")` inside of a node? what is the use case where a user would use both?

No, that shouldn't be done. The logger should only be used "outside" the node. Like in a utils.py module that is imported to be used in the node.

craycrayfish:
> > > > This is a really good feature, but it doesn't feel logical to configure this inside of node config because this is something that would typically be used outside of a node. Do you think there is a good reason not to just have this be a pipeline-level config?
> > > 
> > > 
> > > This allows the user to be very granular in which nodes capture logs or not. But the user can also configure it at the pipeline-level.
> > > > I guess this allows the user to user `print()` inside of a node, but I'm not sure if we want to have both `print()` and `self.log()` available inside a node.
> > > 
> > > 
> > > No, `print` will not be captured.
> > 
> > 
> > @Barabazs Isn't it confusing if the user can use both `self.log()` and `logger = logging.getLogger("foo")` inside of a node? what is the use case where a user would use both?
> 
> No, that shouldn't be done. The logger should only be used "outside" the node. Like in a utils.py module that is imported to be used in the node.

This doesn't seem very user-friendly to me. It would be ideal if users could use the same logging implementation throughout their code, whether in or outside a node. Do you think we could depracate self.log in favor of this?

docs/developer_guide/node_implementation.md
craycrayfish: I'm not sure if this kind of hierarchy makes sense. Since you might have the same utils function being called in multiple nodes but you'll never know which was the node that called the util function and triggered this log.

craycrayfish: > > maybe we can use the NodeManager or one of the internal node's self.log method for this. Or implement an independent logger that doesn't require a self.log
> 
> > If we make it dependent on a `self.log` method call, this will be unable to capture logs that occur before a pipeline goes up and after a pipeline goes down. I think having an independent dataset producer object that writes to the logging topic might be better
> 
> How do you envision this independent object?
> 
> The complexity of logging before/during a pipeline starting lies in the distributed nature of the processes and the creation of the topics. Everything that happens before a node starts isn't visible to the node, so it can't capture and produce. But maybe we can store these logs temporarily until the NodeManager is running and then produce them to the logging topic. It's not a trivial feature, so maybe we should spec that out separately.

I would agree with this IF they turn out to be separate implementations, but I think if we want to stick to one consistent logging mechanism (I think we should) then the more general one, i.e. independent of nodes) should be the one. 

I agree it's more complex - the most straightforward but maybe inefficient method I can think of is for us to create a DatasetProducer in the __init__.py file as a global variable that the `emit` method will use to produce logs. The emit should retry and fail elegantly in cases the logging topic doesn't exist.

Barabazs:
> This doesn't seem very user-friendly to me. It would be ideal if users could use the same logging implementation throughout their code, whether in or outside a node. Do you think we could depracate self.log in favor of this?

I suppose we could do that. But that would be a significant breaking change.

craycrayfish:
> > This doesn't seem very user-friendly to me. It would be ideal if users could use the same logging implementation throughout their code, whether in or outside a node. Do you think we could depracate self.log in favor of this?
> 
> I suppose we could do that. But that would be a significant breaking change.

Let's discuss this next week over a call.

Barabazs:
@craycrayfish  I've been thinking about this PR and I think we should take some time to redesign the entire logging feature and make sure we think about the potential implications and spec out the expected behavior. Maybe for the GA release?

In the meantime I would suggest either:
* merging this feature as-is for the beta users
* or closing this PR

PR #119, opened 2023-12-27
chore: upgrade typehints to py310

Barabazs:
## Summary

Python 3.10 allows for simplified type hints and some type hints were using the older notation.
Example:
`Optional[str]` can be written as `str|None`.

## Motivation

Keeping the type hints consistent across all files.

## Author Checklist

- [x] Changes are motivated and described in the PR description.
- [x] Testing strategy is described if automated tests are not included in the PR.
- [x] Documentation is updated if necessary.
- [x] Status checks pass.
- [x] If the version number is changed, it is updated in `pyproject.toml` , `aineko/__init__.py`, `aineko/templates/first_aineko_pipeline/{{cookiecutter.project_slug}}/pyproject.toml`.

## Reviewer Checklist

- [ ] Title is accurate and formatted appropriately.
- [ ] No unnecessary changes are introduced.
- [ ] Description motivates each change.
- [ ] Docs are updated if necessary.
- [ ] Code is pythonic and consistent with the rest of the codebase.
- [ ] The code is consistent with the [style guide](https://google.github.io/styleguide/pyguide.html).

PR #121, opened 2024-01-04
Feat: WebSocket Connector Node Extra

ricrosales:
## Summary
This PR introduces an extra node for WebSockets. The node is generalized and intended to work with almost any WebSocket data stream.

## Motivation
We'd like to make it easier for Aineko developers to ingest and process data from WebSockets. We've had several use cases that require WebSocket data.

## Author Checklist

- [x] Status checks pass.

## Reviewer Checklist


aineko/extras/connectors/wss.py
```diff
@@ -0,0 +1,160 @@
...
+        # Cast params to ParamsWSS type
+        try:
+            if params is not None:
+                self.wss_params = ParamsWSS(**params)
+            else:
+                raise ValueError(
+                    "No params provided to WSS connector. "
+                    "Node requires at least a url param."
+                )
+        except Exception as err:  # pylint: disable=broad-except
```
marissaposner: I would personally add more specific error handling than just the broad error. Maybe a websocket specific error.

aineko/extras/connectors/wss.py
```diff
@@ -0,0 +1,160 @@
...
+    retry_sleep: float = 5
+    url: str
+    header: Optional[Dict[str, str]] = None
+    init_messages: List[Any] = []
+    metadata: Optional[Dict[str, Any]] = None
+
+    @field_validator("url")
+    def supported_url(cls, u: str) -> str:  # pylint: disable=no-self-argument
+        """Validates that the url is a valid WebSocket URL."""
+        if not u.startswith("wss://") or u.startswith("ws://"):
```
marissaposner: It should be 
`    if not (u.startswith("wss://") or u.startswith("ws://")):
`

marissaposner (commented):
Adding a couple comments

marissaposner (commented):
Would this implementation handle multiple websocket connections simultaneously?

aineko/extras/connectors/wss.py
```diff
@@ -0,0 +1,160 @@
...
+            )
+        return u
+
+
+class WSS(AbstractNode):
+    """Connector for a websocket."""
+
+    retry_count = 0
+
+    def _pre_loop_hook(self, params: dict | None = None) -> None:
```
mdigi14:
```suggestion
    def _pre_loop_hook(self, params: Optional[dict] = None) -> None:
```

I believe this is the syntax we should use based on our style conventions.

mdigi14:
@ricrosales can you add more context around how the secrets injection works? I'm not sure what the overarching goal is and why there are multiple helper functions for different types.

aineko/extras/connectors/wss.py
ricrosales: This logic is necessary since ray cannot pickle all exceptions. So the workaround is to catch all exceptions and raise an exception that ray can pickle, while casting the error to a str so you don't lose information about the error. in this case the possible excpetions are pydantic exceptions, which are known to be non-picklable.

I'll add some documentation for this.

ricrosales: Documented here: https://github.com/aineko-dev/aineko/pull/121/commits/faf2d9754b685ce6a87e0b10380a9d6abef8e481

aineko/extras/connectors/wss.py
ricrosales: good catch, fixed here: https://github.com/aineko-dev/aineko/pull/121/commits/50c450677ff40e7d1efc554fb6bdd128c59136d9

aineko/extras/connectors/wss.py
ricrosales: I did this because I was failing mypy checks locally, but i think i had some stale settings. i fixed this here: https://github.com/aineko-dev/aineko/pull/121/commits/a4db5b4f648ceec3478a51d71eaff97b55383ac2

ricrosales:
> @ricrosales can you add more context around how the secrets injection works? I'm not sure what the overarching goal is and why there are multiple helper functions for different types.

The `inject_secrets` method is used by an aineko node (`WSS`, in this case) to inject secrets into objects passed in to node params via a pipeline config.

Given an object, inject secrets from the environment into the object recursively. Secrets are identified in strings by the pattern {$SECRET_NAME} where SECRET_NAME is the name of the environment variable to inject. For example, given the following environment variables:

```
$ export SECRET1=secret1
$ export SECRET2=secret2
```

The following string:

```
"This is a string with a {$SECRET1} and a {$SECRET2}."
```

Would be transformed to:

```
"This is a string with a secret1 and a secret2."
```

I needed to support secret injection into list, str, and dict for WSS so that i could provide auth keys/secrets using various auth formats utilized by WebSockets. Authenticating via a message requires list, dict, and string injection. Authenticating via URL only requires str injection. etc...

I added a docstring with this context here: https://github.com/aineko-dev/aineko/pull/121/commits/eb7cc52fad7aea02aec07bf6803ec523e062aef5

Also noted this behavior directly in the docstring here: https://github.com/aineko-dev/aineko/pull/121/commits/1e97a34bcde8e66a8e8cc508808dbfa7c886556e

ricrosales:
@craycrayfish any thoughts on how best to write a test for this node?

ricrosales:
Also @craycrayfish / @giddyphysicist any thoughts on the naming convention for the node and the connector directory? I will follow this PR with a PR for a HTTPS connector, for reference.

aineko/extras/connectors/wss.py
Barabazs: No the original code was correct. We're now using python 3.10 type hints https://github.com/aineko-dev/aineko/pull/119

@ricrosales Can you change it back to the correct py310 syntax please?

Barabazs (commented):
I'm not a big fan of naming this "wss". Afaik this is an acronym for the protocol "WebSocket Secure" and not an acronym for "websocket".

aineko/utils/secrets.py
Barabazs, line 1: This should probably be moved to `extras/connectors/...`

docs/extras/wss.md
```diff
@@ -0,0 +1,13 @@
+# WSS (WebSocket) Node
+
+The WSS extra node can be used by adding the following to `pyproject.toml`
+
+:   
+    ```yaml title="pyproject.toml" hl_lines="2"
+    [tool.poetry.dependencies]
+    aineko = {version = "^0.2.7", extras=["wss"]}
+    ```
```
Barabazs: It might be better to tell the user to do `poetry add aineko[wss]`.
That command will take care of updating the pyproject.toml, updating the poetry.lock and installing the new dependencies.

aineko/extras/connectors/wss.py
```diff
@@ -0,0 +1,209 @@
...
+
+    max_retries: int = 30
+    retry_sleep: float = 5
+    url: str
+    header: Optional[Dict[str, str]] = None
+    init_messages: List[Any] = []
+    metadata: Optional[Dict[str, Any]] = None
+
+    @field_validator("url")
+    def supported_url(cls, u: str) -> str:  # pylint: disable=no-self-argument
```
Barabazs:
```suggestion
    @field_validator("url")
    @classmethod
    def supported_url(cls, u: str) -> str:
```
This should be decorated as a classmethod.

Barabazs: Single character names should be avoided.

aineko/extras/connectors/wss.py
```diff
@@ -0,0 +1,209 @@
...
+                )
+                self.retry_count += 1
+                time.sleep(self.wss_params.retry_sleep)
+                self.create_subscription()
+            else:
+                raise ValueError(
+                    "Retry count exceeded max retries. "
+                    "Failed to create subscription to "
+                    f"{str(self.wss_params.url)}. "
+                    f"The following error occured: {str(err)}"
```
Barabazs:
```suggestion
                    f"The following error occurred: {err}"
```
typo + redundant type casting

aineko/extras/connectors/wss.py
```diff
@@ -0,0 +1,209 @@
...
+    metadata: Optional[Dict[str, Any]] = None
+
+    @field_validator("url")
+    def supported_url(cls, u: str) -> str:  # pylint: disable=no-self-argument
+        """Validates that the url is a valid WebSocket URL."""
+        if not (u.startswith("wss://") or u.startswith("ws://")):
+            raise ValueError(
+                "Invalid url provided to WebSocket params. "
+                'Expected url to start with "wss://" or "ws://". '
+                f"Provided url was: {str(u)}"
```
Barabazs:
```suggestion
                f"Provided url was: {u}"
```
The str() is redundant when using f-strings.

I noticed this pattern in the rest of the code as well, but I'll only comment on this instance fyi.

pyproject.toml
```diff
@@ -32,6 +32,8 @@ pygithub = "~=2.1.1"
 # extras: fastapi
 fastapi = { version = ">0.101", optional = true }
 uvicorn = { version = "^0.22", extras = ["standard"], optional = true }
+# extras: wss
+websocket-client = { version = "^1.7.0", optional = true }
```
Barabazs, line 36: Is there a specific reason you chose this package?

Barabazs (changes requested):
Intermediate review with some change requests;

ricrosales:
> Would this implementation handle multiple websocket connections simultaneously?

In this implementation, multiple websockets would be handled via multiple instances of the WSS node. I think this is more of an Aineko-esque approach given that it's easy to manage many separate nodes. This also means that errors in different websockets won't interact.

aineko/extras/connectors/wss.py
ricrosales: reverted here: https://github.com/aineko-dev/aineko/pull/121/commits/86629c74f4be340b4cfa42a0df907f348ea628b6

Thanks @Barabazs

ricrosales:
> I'm not a big fan of naming this "wss". Afaik this is an acronym for the protocol "WebSocket Secure" and not an acronym for "websocket".

I liked it previously, because i thought i would only support wss, but this node supports both wss and ws protocols. I didn't have high conviction on the naming, so i'll switch it to WebSocket. I originally did not want to use WebSocket because that felt like a name collision with other websocket implementations.

aineko/utils/secrets.py
ricrosales: makes sense. moved here: https://github.com/aineko-dev/aineko/pull/121/commits/034214773a4e436fb78b9003a1a9b70724768d31

docs/extras/wss.md
ricrosales: @craycrayfish any reason you did not do it the way @Barabazs suggests for FastAPI?

aineko/extras/connectors/wss.py
ricrosales: fixed here: https://github.com/aineko-dev/aineko/pull/121/commits/184212cfcd0372da8112b042fe282c001ac8cc5f

note: i referenced the models directory in the codebase. maybe we should update all of these? 
e.g. https://github.com/aineko-dev/aineko/blob/develop/aineko/models/deploy_config_schema.py#L32-L37

ricrosales: fixed here: https://github.com/aineko-dev/aineko/pull/121/commits/7eb4998eee7e29d284e76097c5f12d44c84e7e78

aineko/extras/connectors/wss.py
ricrosales: fixed here: https://github.com/aineko-dev/aineko/pull/121/commits/d71cd6b4c6b3f85bd4230ef93074e42ece1fdf78

aineko/extras/connectors/wss.py
ricrosales: oh i mis-interpreted a mypy error i saw. fixed here: https://github.com/aineko-dev/aineko/pull/121/commits/d71cd6b4c6b3f85bd4230ef93074e42ece1fdf78

pyproject.toml
ricrosales: This was a package we chose a while back in a specific pipeline implementation and it has worked well since then. 

The project looks decently well maintained and has no dependencies. 

After more research it seems like the `websocket-client` library is probably a better choice for us over the other well maintained `websockets` library given it allows flexibility to work with both synchronous and asynchronous scenarios. A major benefit for `websockets` is that there is less boilerplate for managing simultaneous connections. In our scenario, we'd manage simultaneous connections across Aineko nodes, so that is not a concern.

aineko/extras/connectors/websocket.py
```diff
@@ -0,0 +1,210 @@
...
+    def _execute(self, params: dict | None = None) -> None:
+        """Polls and gets data from the WebSocket."""
+        try:
+            # Poll the websocket
+            raw_message = self.ws.recv()
+        except Exception as err:  # pylint: disable=broad-except
+            # If the connection is closed, reconnect
+            self.log(
+                "Websocket connection closed. "
+                f"Reconnecting in {self.ws_params.retry_sleep} "
```
craycrayfish: We should use an exponential back off retry strategy by default.

craycrayfish:
> @craycrayfish any thoughts on how best to write a test for this node?

I think we would need to set up a fake websocket as part of the test, then use the node to make sure it connects. You can refer to the integration test for the FastAPI node for an example.

craycrayfish:
> @craycrayfish any thoughts on how best to write a test for this node?

I propose we have a subdirectory of extras called api. So extras.api.websocket and later extras.api.rest

aineko/extras/connectors/wss.py
Barabazs: Good point. Let's do that in a separate PR to keep this one focused on your new feature.

Barabazs: Addressed in https://github.com/aineko-dev/aineko/pull/125

docs/extras/wss.md
craycrayfish: I think both ways are fine. We can use @Barabazs way in the docs and power users will figure out the other method by themselves if they know.

aineko/extras/connectors/websocket.py
```diff
@@ -0,0 +1,210 @@
...
+from pydantic import BaseModel, field_validator
+
+from aineko import AbstractNode
+from aineko.extras.connectors.secrets import inject_secrets
+
+
+class ParamsWebSocket(BaseModel):
+    """Connector params for WebSocket model."""
+
+    max_retries: int = 30
```
craycrayfish: This should perhaps be -1 by default so it keeps trying indefinitely by default

Barabazs:
> I propose we have a subdirectory of extras called api. So extras.api.websocket and later extras.api.rest

API seems like a better group name for our FastAPI node tbh.
How about something like:
* data_handlers
* data_pollers
* data_fetchers

pyproject.toml
Barabazs: Thanks, I appreciate the well thought out response. 🙏

aineko/extras/connectors/websocket.py
```diff
@@ -0,0 +1,210 @@
...
+                self.log(
+                    "Encountered error when attempting to connect to "
+                    f"{self.ws_params.url}. Will retry in "
+                    f"{self.ws_params.retry_sleep} seconds"
+                )
+                self.retry_count += 1
+                time.sleep(self.ws_params.retry_sleep)
+                self.create_subscription()
+            else:
+                raise ValueError(
```
Barabazs: This should be mentioned in the docstring as

```
Raises:
    ValueError: xxxx
```

aineko/extras/connectors/websocket.py
```diff
@@ -0,0 +1,210 @@
...
+                    f"Failed to parse message: {raw_message!r}. "
+                    f"The following error occured: {err} "
+                    f"Reconnecting in {self.ws_params.retry_sleep} "
+                    "seconds...",
+                    level="error",
+                )
+                time.sleep(self.ws_params.retry_sleep)
+                self.create_subscription()
+            else:
+                raise ValueError(
```
Barabazs: Can you also document this in the docstring please

aineko/extras/connectors/secrets.py
Barabazs, line 1: Can you also upgrade the type hints here please?
dict[str, Any] instead of Dict[str, Any] etc.

aineko/extras/connectors/secrets.py
```diff
@@ -0,0 +1,95 @@
...
+        if isinstance(v, str):
+            list_[i] = _str_inject_secrets(v)
+        elif isinstance(v, dict):
+            list_[i] = _dict_inject_secrets(v)
+        elif isinstance(v, list):
+            list_[i] = _list_inject_secrets(v)
+    return list_
+
+
+def inject_secrets(obj: Any) -> Any:
```
Barabazs: I don't think we should be using Any here.

aineko/extras/connectors/websocket.py
```diff
@@ -0,0 +1,210 @@
...
+
+class ParamsWebSocket(BaseModel):
+    """Connector params for WebSocket model."""
+
+    max_retries: int = 30
+    retry_sleep: float = 5
+    url: str
+    header: Optional[Dict[str, str]] = None
+    init_messages: List[Any] = []
+    metadata: Optional[Dict[str, Any]] = None
```
Barabazs:
```suggestion
    header: dict[str, str] | None = None
    init_messages: list[Any] = []
    metadata: dict[str, Any] | None = None
```

Barabazs (changes requested):
Can you also add tests for the secrets.py and the websocket node please?

aineko/extras/connectors/secrets.py
mdigi14, line 1: I think the syntax in this file needs to be refactored. This code is not very readable imo.

Examples:

- Parameter names in the functions like `str_`, `dict_`, `list_`
- Function names like `_str_inject_secrets`

aineko/extras/connectors/secrets.py
```diff
@@ -0,0 +1,95 @@
+# Copyright 2023 Aineko Authors
+# SPDX-License-Identifier: Apache-2.0
+"""Secrets utilities."""
+
+import os
+import re
+from typing import Any, Dict, List
+
+
+def _str_inject_secrets(str_: str) -> str:
```
mdigi14: It would be helpful to explain what is happening here with the regex and recursion

aineko/extras/connectors/secrets.py
```diff
@@ -0,0 +1,95 @@
...
+    """Inject secrets from environment into a str."""
+    secret_pattern = r"\{\$.*?\}"
+    secret_match = re.search(secret_pattern, str_, re.DOTALL)
+    if not secret_match:
```
mdigi14: Also you can remove the `else` block since you are returning

craycrayfish:
I've been thinking about the secret injection and I think it might be better to add it to the config loader so secrets can be used anywhere in the config.

aineko/extras/connectors/secrets.py
ricrosales: Do you have a proposal for a change? the function and variable names seem descriptive to me.

aineko/extras/connectors/secrets.py
ricrosales: added more documentation

ricrosales:
> I've been thinking about the secret injection and I think it might be better to add it to the config loader so secrets can be used anywhere in the config.

Good idea. This is done.

aineko/core/config_loader.py
```diff
@@ -62,6 +65,11 @@ def load_config(self) -> dict:
             )
             raise e
 
+        # Inject environment variables into node params
+        for node in config["pipeline"]["nodes"].values():
+            if node["node_params"] is not None:
```
craycrayfish: I think we moved the convention to 
```
if node["node_params"]:
```

aineko/extras/connectors/websocket_client.py
```diff
@@ -0,0 +1,216 @@
...
+            time.sleep(self.ws_params.retry_sleep)
+            self.create_subscription()
+            return
+
+        try:
+            # Parse the message and emit to producers
+            message = json.loads(raw_message)
+            if self.ws_params.metadata is not None:
+                message = {
+                    "metadata": self.ws_params.metadata,
```
giddyphysicist, line 153: Here, it looks like the same metadata would be added to every message from this connection. Would there be a case for adding different metadata per message type, or based on the added complexity do you think this would be better handled by a separate data processing node?

aineko/extras/connectors/websocket_client.py
```diff
@@ -0,0 +1,216 @@
...
+                    "data": message,
+                }
+            for dataset, producer in self.producers.items():
+                if dataset != "logging":
+                    producer.produce(message)
```
giddyphysicist: Similar to the comment above, but here it looks like you could give the websocket any number of producer topics that would have duplicate data from all messages. Would it be worth routing before that, or is the goal here to capture just the raw websocket data which could be processed later?

aineko/extras/connectors/websocket_client.py
ricrosales: This is intended to help add context to the raw websocket messages that make it easier to parse in downstream nodes. i think i understand what you are suggesting, but i cant think of a good way to do this type of complex metadata injection in a way that in generalized enough for this node.

aineko/extras/connectors/websocket_client.py
ricrosales: i'm not a big fan of doing any sort of routing in this node. this node is intended to strictly speaking write raw data from the websocket to be processed downstream. i did this here to handle scenarios where the user sets more than one output dataset. since we are writing the same data and copying shouldn't be useful, i'll update this logic to validate that there is only one output dataset.

ricrosales:
> Can you also add tests for the secrets.py and the websocket node please?

this is done

ricrosales:
I believe I've addressed all the comments here. Can we please move to merge this?

craycrayfish:
Thank you for doing this. Will approve, but be aware that #134 will require typehint changes if this is merged first.

craycrayfish (approved):

giddyphysicist (approved):

tests/core/test_config_loader.py
```diff
@@ -48,14 +53,26 @@
...
+    monkeypatch.setenv("AINEKO_TEST_LIST_VAR_2", "two")
+    monkeypatch.setenv("AINEKO_TEST_LIST_VAR_3", "three")
+    monkeypatch.setenv("AINEKO_TEST_DICT_VAR_1", "value_1")
+    monkeypatch.setenv("AINEKO_TEST_DICT_VAR_2", "value_2")
+
     # Test pipeline config containing single pipeline
     config = ConfigLoader(test_pipeline_config_file).load_config()
+    print(config)
+    print("EXPECTED")
+    print(EXPECTED_TEST_PIPELINE)
```
Barabazs: Can you remove this if it's not needed for the tests?

Barabazs (changes requested):
Just one small change request 😺

aineko/extras/connectors/websocket.py
ricrosales: I have a list of future improvements. This is one of them. Will leave to implement in a follow on.

Barabazs (approved):
:fire:

PR #124, opened 2024-01-06
Feat: HTTP Connector Node Extra

ricrosales:
## Summary
This PR introduces an extra node for HTTP API polling. The node is generalized and intended to work with almost any HTTP API using GET.

I added tests here: https://github.com/ricrosales/wss-test.

Docs updates are incoming.

This PR builds on [feat/wss-connector](https://github.com/aineko-dev/aineko/tree/feat/wss-connector), so merging there first.

## Motivation
We'd like to make it easier for Aineko developers to ingest and process data from HTTP APIs. We've had several use cases that require HTTP API data.

## Author Checklist

- [x] Status checks pass.

## Reviewer Checklist


aineko/extras/connectors/rest.py
```diff
@@ -0,0 +1,152 @@
...
+
+    timeout: int = 10
+    url: str
+    headers: Optional[Dict[str, Any]] = None
+    data: Optional[Dict[str, Any]] = None
+    poll_interval: int = 5
+    poll_throttle: float = 0.1
+    max_retries: int = 30
+    metadata: Optional[Dict[str, Any]] = None
+    retry_sleep: float = 5
```
craycrayfish: This should be an exponential backoff as well

aineko/extras/connectors/rest.py
```diff
@@ -0,0 +1,152 @@
...
+            # Note: this is required because pydantic errors
+            # are not pickleable
+            raise ValueError(
+                "Failed to cast params to ParamsREST type. "
+                f"The following error occurred: {err}"
+            ) from err
+        self.rest_params.headers = inject_secrets(self.rest_params.headers)
+        self.rest_params.url = inject_secrets(self.rest_params.url)
+        self.rest_params.data = inject_secrets(self.rest_params.data)
+
```
craycrayfish: Instead of injecting secrets here, we can inject them in the pydantic model by running the function for the relevant fields before enforcing checks.

aineko/extras/connectors/rest.py
```diff
@@ -0,0 +1,152 @@
...
+            try:
+                # Poll REST api
+                response = self.session.get(
+                    self.rest_params.url,
+                    timeout=self.rest_params.timeout,
+                    headers=self.rest_params.headers,
+                    data=self.rest_params.data,
+                )
+                # Check if the request was successful
+                if response.status_code != 200:
```
craycrayfish: Should we broaden the scope to none 2XX codes. I think 2XX codes are all success

aineko/extras/connectors/rest.py
```diff
@@ -0,0 +1,152 @@
...
+            raise ValueError(
+                "Invalid url provided to HTTPS params. "
+                'Expected url to start with "https://". '
+                f"Provided url was: {url}"
+            )
+        return url
+
+
+class REST(AbstractNode):
+    """Connects to an REST endpoint via HTTPS."""
```
craycrayfish: Can you please include an example of how to configure such a node in a yml config file? As well as default behaviors such as retry patterns, etc

aineko/extras/connectors/rest.py
Barabazs: +1

aineko/extras/connectors/rest.py
```diff
@@ -0,0 +1,152 @@
...
+    @classmethod
+    def supported_url(cls, url: str) -> str:  # pylint: disable=no-self-argument
+        """Validates that the url is a valid HTTPS URL."""
+        if not (url.startswith("https://") or url.startswith("http://")):
+            raise ValueError(
+                "Invalid url provided to HTTPS params. "
+                'Expected url to start with "https://". '
+                f"Provided url was: {url}"
+            )
+        return url
```
Barabazs: Why should we not support http?

aineko/extras/connectors/rest.py
Barabazs, line 27:
```suggestion
    headers: dict[str, Any] | None = None
    data: dict[str, Any] | None = None
    poll_interval: int = 5
    poll_throttle: float = 0.1
    max_retries: int = 30
    metadata: dict[str, Any] | None = None
    retry_sleep: float = 5
```

Barabazs: Can you document these fields please?

Barabazs (changes requested):

aineko/extras/connectors/rest.py
```diff
@@ -0,0 +1,152 @@
...
+        # Create a session
+        self.log(
+            f"Creating new session to REST endpoint {self.rest_params.url}."
+        )
+        self.session = requests.Session()
+
+    def _execute(self, params: dict | None = None) -> None:
+        """Polls and gets data from the HTTPS endpoint."""
+        # Check if it is time to poll
+        if time.time() - self.last_poll_time >= self.rest_params.poll_interval:
```
mdigi14: The relationship between `poll_interval` and `poll_throttle` isn't clear to me. Can you add some more documentation

aineko/extras/connectors/rest.py
```diff
@@ -0,0 +1,152 @@
...
+    """Connector params for REST model."""
+
+    timeout: int = 10
+    url: str
+    headers: Optional[Dict[str, Any]] = None
+    data: Optional[Dict[str, Any]] = None
+    poll_interval: int = 5
+    poll_throttle: float = 0.1
+    max_retries: int = 30
+    metadata: Optional[Dict[str, Any]] = None
```
mdigi14: It's not clear how `metadata` is supposed to be used.

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,158 @@
...
+    poll_interval: int = 5
+    poll_throttle: float = 0.1
+    max_retries: int = 30
+    metadata: Optional[Dict[str, Any]] = None
+    retry_sleep: float = 5
+
+    @field_validator("url")
+    @classmethod
+    def supported_url(cls, url: str) -> str:  # pylint: disable=no-self-argument
+        """Validates that the url is a valid HTTPS URL."""
```
giddyphysicist: small docstring update to include http:
`Validates that the url is a valid HTTP or HTTPS URL`

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,158 @@
...
+            raise ValueError(
+                "Invalid url provided to HTTPS params. "
+                'Expected url to start with "https://". '
+                f"Provided url was: {url}"
+            )
+        return url
+
+
+class RESTPoller(AbstractNode):
+    """Connects to an REST endpoint via HTTPS and polls."""
```
giddyphysicist: similar to above -- adding http to the docstring:
`Connects to an REST endpoint via HTTP or HTTPS and polls.`

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,158 @@
...
+        self.output_dataset = output_datasets[0]
+
+        # Create a session
+        self.log(
+            f"Creating new session to REST endpoint {self.rest_params.url}."
+        )
+        self.session = requests.Session()
+
+    def _execute(self, params: dict | None = None) -> None:
+        """Polls and gets data from the HTTPS endpoint."""
```
giddyphysicist: docstring update: 
`Polls and gets data from the HTTP or HTTPS endpoint.`

aineko/extras/connectors/rest.py
giddyphysicist: would it be worth adding an optional set of allowable expected return codes, which defaults to 2XX?

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,158 @@
+# Copyright 2023 Aineko Authors
+# SPDX-License-Identifier: Apache-2.0
+"""Extra module for connecting to a REST endpoint."""
+
+import json
+import time
+from typing import Any, Dict, Optional
+
+import requests
```
giddyphysicist: have you considered using httpx (or some other async non-blocking client) rather than requests library?

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # Ensure only one output dataset is provided
+        output_datasets = [d for d in self.producers.keys() if d != "logging"]
+        if len(output_datasets) > 1:
+            raise ValueError(
+                "Only one output dataset is allowed for the "
+                "RESTPoller connector. "
+                f"{len(output_datasets)} datasets given."
+            )
+        self.output_dataset = output_datasets[0]
```
mdigi14: `self.output_dataset` is defined but not used anywhere

mdigi14 (changes requested):
Minor comment related to unused class field

pyproject.toml
```diff
@@ -35,6 +35,9 @@ uvicorn = { version = "^0.22", extras = ["standard"], optional = true }
 # extras: websocket_client
 websocket-client = { version = "^1.7.0", optional = true }
 websockets = { version = "^12.0", optional = true }
+# extras: rest
+requests = { version = "^2.31.0", optional = true }
+types-requests = { version = "^2.31.0", optional = true }
```
Barabazs:
```suggestion
# extras: rest
requests = { version = "^2.31.0", optional = true }
```
Installing this is handled by mypy when needed. I don't think we should add unneeded packages by default.

pyproject.toml
```diff
@@ -66,6 +69,7 @@ pytest-subtests = "^0.11.0"
 
 [tool.poetry.extras]
 fastapi = ["fastapi", "uvicorn"]
+rest = ["requests", "types-requests"]
```
Barabazs:
```suggestion
rest = ["requests"]
```

docs/extras/websocket.md
Barabazs, line 1: This seems like a merge error.

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,158 @@
...
+    data: Optional[Dict[str, Any]] = None
+    poll_interval: int = 5
+    poll_throttle: float = 0.1
+    max_retries: int = 30
+    metadata: Optional[Dict[str, Any]] = None
+    retry_sleep: float = 5
+
+    @field_validator("url")
+    @classmethod
+    def supported_url(cls, url: str) -> str:  # pylint: disable=no-self-argument
```
Barabazs:
```suggestion
    def supported_url(cls, url: str) -> str:
```
Not needed when adding the `@classmethod` decorator.

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,158 @@
...
+            # so that it can be properly caught by the node
+            # Note: this is required because pydantic errors
+            # are not pickleable
+            raise ValueError(
+                "Failed to cast params to ParamsREST type. "
+                f"The following error occurred: {err}"
+            ) from err
+
+        # Ensure only one output dataset is provided
+        output_datasets = [d for d in self.producers.keys() if d != "logging"]
```
Barabazs: Single character names should be avoided if we want to stick with the [styleguide](https://google.github.io/styleguide/pyguide.html#3161-names-to-avoid).

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,158 @@
...
+                else:
+                    raise ValueError(
+                        "Retry count exceeded max retries "
+                        f"({self.rest_params.max_retries}). "
+                        f"Failed to parse message: {raw_message}. "
+                        f"The following error occurred: {err}"
+                    ) from err
+        else:
+            # If it is not time to poll, sleep
+            time.sleep(self.rest_params.poll_throttle)
```
Barabazs: Wouldn't it be better to sleep for only the time remaining until ~ `time.time() - self.last_poll_time >= self.rest_params.poll_interval`

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,158 @@
...
+                    self.log(
+                        f"Failed to parse message: {raw_message}. "
+                        f"The following error occurred: {err} "
+                        f"Will retry in {self.rest_params.retry_sleep} "
+                        "seconds...",
+                        level="error",
+                    )
+                    time.sleep(self.rest_params.retry_sleep)
+                else:
+                    raise ValueError(
```
Barabazs: Using a ValueError in this case is confusing imo.

Barabazs (changes requested):

Barabazs:
@ricrosales Is it this PR actually ready for review? If not, can you mark it as draft please?

aineko/extras/connectors/rest.py
ricrosales: Will do this in a later PR

aineko/extras/connectors/rest.py
ricrosales: We're doing this via the config loader now.

aineko/extras/connectors/rest.py
ricrosales: Great idea. Updated.

aineko/extras/connectors/rest.py
ricrosales: Yes I will document similar to WebSocketClient

aineko/extras/connectors/rest.py
ricrosales: Yes will do same as with WebSocketClient.

aineko/extras/connectors/rest.py
ricrosales: Will document like in the WebSocketClient.

aineko/extras/connectors/rest_poller.py
ricrosales: fixed

pyproject.toml
ricrosales: this is a holdover from me not properly installing with dev,test groups. fixed.

docs/extras/websocket.md
ricrosales: agreed. deleted.

aineko/extras/connectors/rest_poller.py
ricrosales: got it, fixed

aineko/extras/connectors/rest_poller.py
ricrosales: fixed here and for WebsocketClient

aineko/extras/connectors/rest_poller.py
ricrosales: yeah... not sure what i was thinking before. updating with this logic:
```
        else:
            # If it is not time to poll, sleep
            time.sleep(
                self.rest_params.poll_interval
                - (time.time() - self.last_poll_time)
            )
```

aineko/extras/connectors/rest_poller.py
ricrosales: In the absence of a custom error, I'll switch this to be a broad exception.

ricrosales:
> @ricrosales Is it this PR actually ready for review? If not, can you mark it as draft please?

My fault, I don't recall converting this to review. This is not ready. I need to add documentation and add tests after merging in upstream develop changes.

Marked as draft for now. Will get this out asap.

aineko/extras/connectors/rest_poller.py
ricrosales: Good question. Correct me I'm wrong but in the particular case of this node, blocking seems ok to me since we are "waiting" for data to come back to the api in order to put them into the dataset. i think this is ok since we are polling on an interval as opposed to trying to fetch data as fast as possible.

Barabazs: I agree with Ricardo's reasoning. Sync requests should be fine in this case.

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+# pylint: disable=anomalous-backslash-in-string
+class RESTPoller(AbstractNode):
+    """Connects to an REST endpoint via HTTP or HTTPS and polls.
+
+    This node is a wrapper around the
+    [requests](
+        https://docs.python-requests.org/en/master/
+        ){:target="\_blank"} library.
```
Barabazs:
```suggestion
class RESTPoller(AbstractNode):
    """Connects to an REST endpoint via HTTP or HTTPS and polls.

    This node is a wrapper around the
    [requests](
        https://docs.python-requests.org/en/master/
        ){:target="_blank"} library.
```
I don't think this backslash is needed to get the expected behavior. See PR #141

docs/extras/rest_poller.md
```diff
@@ -0,0 +1,13 @@
...
+
+:   
+    ```yaml title="pyproject.toml" hl_lines="2"
+    [tool.poetry.dependencies]
+    aineko = {version = "^0.2.7", extras=["rest_poller"]}
+    ```
+
+## API Reference
+
+::: aineko.extras.connectors.rest_poller.RestPoller
```
Barabazs:
```suggestion
::: aineko.extras.connectors.rest_poller.RESTPoller
```
Need to use the exact class name or it won't be included in the docs.

docs/extras/rest_poller.md
Barabazs, line 1: Also need to add this to the mkdocs.yml nav:
```yaml
          - extras/rest_poller.md
```

tests/extras/connectors/main.py
Barabazs, line 1: Maybe this file should be renamed to something more descriptive?

tests/extras/connectors/test_rest_poller.py
```diff
@@ -0,0 +1,55 @@
...
+            has_pipeline_prefix=True,
+        )
+        test_results = consumer.next()
+        assert test_results["message"] == {
+            "message_0": "Hello World!",
+            "message_1": "Hello World!",
+            "message_2": "Hello World!",
+            "message_3": "Hello World!",
+            "message_4": "Hello World!",
+        }
```
Barabazs: imo this needs more tests. This only tests one specific and limited behavior of the node.
Maybe include some tests on how it reacts to certain status codes, timeouts, network drops, different types of response content,...

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,260 @@
...
+        # Ensure only one output dataset is provided
+        output_datasets = [
+            dataset for dataset in self.producers.keys() if dataset != "logging"
+        ]
+        if len(output_datasets) > 1:
+            raise ValueError(
+                "Only one output dataset is allowed for the "
+                "RESTPoller connector. "
+                f"{len(output_datasets)} datasets given."
+            )
```
Barabazs: I don't think this is documented in the docstrings, right? I would definitely include that in the docs.

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,260 @@
...
+    `timeout` to prevent the node from timing out.
+    """
+
+    # Poll settings
+    last_poll_time = time.time()
+    retry_count = 0
+
+    def _pre_loop_hook(self, params: Optional[Dict] = None) -> None:
+        """Initializes connection to API."""
+        # Cast params to ParamsREST type
```
Barabazs: Personal opinion: comments like this don't add any value

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,260 @@
...
+                polls. Defaults to 5.
+            max_retries (optional): The maximum number of times to retry
+                connecting to the REST endpoint. Defaults to -1.
+            retry_sleep (optional): The number of seconds to wait between
+                retries. Defaults to 5.
+            metadata (optional): A dictionary of metadata to attach to outgoing
+                messages. Defaults to None.
+            success_codes (optional): A list of HTTP status codes that indicate
+                success. Defaults to
+                    [200, 201, 202, 203, 204, 205, 206, 207, 208, 226].
```
Barabazs: Is there a reason why you prefer to document this here instead of properly documenting the Pydantic model and referencing it here + include it in the docs? That would be more maintainable imo and might improve readability.

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,260 @@
...
+
+    By default, this node will not attach headers or data to the request. This
+    can be changed by setting the `headers` and `data` fields in `node_params`.
+    If the REST endpoint requires headers or data to be sent, it is
+    you should set the `headers` and `data` fields in `node_params`.
+
+    By default, this node will timeout after 10 seconds. This can be changed
+    by setting the `timeout` field in `node_params`. If the REST endpoint is
+    expected to take a long time to respond, it is recommended to increase
+    `timeout` to prevent the node from timing out.
```
Barabazs: I wonder if this text might be more concise and less repetitive if we document the behavior of the settings by documenting the settings themselves, as mentioned in my previous comment.

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,260 @@
+# Copyright 2023 Aineko Authors
+# SPDX-License-Identifier: Apache-2.0
+"""Extra module for connecting to a REST endpoint."""
```
Barabazs: Maybe we should document "why" a user might want to use this node. I don't think it's mentioned anywhere what the use case is and how it might benefit users.

Barabazs (dismissed):

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,260 @@
...
+    that contains the value `SECRET_VALUE`, you can inject it into the url
+    field by passing `https://example.com?secret={$SECRET_NAME}` as the url.
+    The connector will then replace `{$SECRET_NAME}` with `SECRET_VALUE` before
+    connecting to the REST endpoint.
+
+    Example usage in pipeline.yml:
+    ```yaml title="pipeline.yml"
+    pipeline:
+      nodes:
+        RestPoller:
```
mdigi14: The class is named `RESTPoller`. There is another incorrect reference a few lines below

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,260 @@
...
+                connecting to the REST endpoint. Defaults to -1.
+            retry_sleep (optional): The number of seconds to wait between
+                retries. Defaults to 5.
+            metadata (optional): A dictionary of metadata to attach to outgoing
+                messages. Defaults to None.
+            success_codes (optional): A list of HTTP status codes that indicate
+                success. Defaults to
+                    [200, 201, 202, 203, 204, 205, 206, 207, 208, 226].
+
+    Secrets can be injected (from environment) into the `url`, `headers`, and
```
mdigi14: I don't think this is that clear for beginner users. Are you suggesting to just inject secrets from a `.env` and then to reference them using the $ syntax?

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,260 @@
...
+                self.log(
+                    "Request failed. "
+                    f"Sleeping for {self.rest_params.poll_interval} seconds. "
+                    f"Error: {err}",
+                    level="error",
+                )
+                time.sleep(self.rest_params.poll_interval)
+                self.retry_count += 1
+                # Reset the session
+                self.log(
```
mdigi14: Why are we resetting the session? Are we just assuming that resetting the session might resolve whatever error is caught here?

aineko/extras/connectors/rest_poller.py
mdigi14: I could be wrong, but I'm pretty sure we're making a mistake to label anything here as `REST` based. Nothing here is specifically labeled to REST and we could also make GraphQL calls using the `requests` library.

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,260 @@
...
+            Exception: If the retry count exceeds the max retries.
+        """
+        # Check if it is time to poll
+        if time.time() - self.last_poll_time >= self.rest_params.poll_interval:
+            # Update the last poll time
+            self.last_poll_time = time.time()
+
+            try:
+                # Poll REST api
+                response = self.session.get(
```
mdigi14: We might want to expose `params` and `json` and `auth` params in addition to `data`

aineko/extras/connectors/rest_poller.py
```diff
@@ -0,0 +1,260 @@
...
+                else:
+                    raise Exception(  # pylint: disable=broad-exception-raised
+                        "Retry count exceeded max retries "
+                        f"({self.rest_params.max_retries}). "
+                        f"Failed to parse message: {raw_message}. "
+                        f"The following error occurred: {err}"
+                    ) from err
+        else:
+            # If it is not time to poll, sleep
+            time.sleep(
```
mdigi14: I'm pretty sure this number could be negative and the `time` library would raise a `ValueError`.

Example:
`self.rest_params.poll_interval` = 0.1
`time.time()` = 1706198047
`self.last_poll_time` = 1706198045 (implies the API call took roughly 2 seconds)

mdigi14 (dismissed):
Caught one bug, then some minor questions / style stuff. Also a comment about exposing some other features in the `requests` lib

aineko/extras/connectors/rest_poller.py
mdigi14: The result of this math would be -4.9

docs/extras/rest_poller.md
ricrosales: done

tests/extras/connectors/main.py
ricrosales: agreed and done

tests/extras/connectors/test_rest_poller.py
ricrosales: i agree, but i'll do this as a follow up. same is true for websocket node.

aineko/extras/connectors/rest_poller.py
ricrosales: done

aineko/extras/connectors/rest_poller.py
ricrosales: sure.  i tend to like to write comments such that 6 months from now i can scan comments and understand the functionality without fully understanding the code. since we're not really following this pattern everywhere, im fine removing this.

aineko/extras/connectors/rest_poller.py
ricrosales: mostly because i don't know how to do what you are describing. is there an example i can look at?

aineko/extras/connectors/rest_poller.py
ricrosales: I added these comments to address shawn's point about documenting default behavior. i thought it'd be useful to be as explicit about that as possible, as opposed to having the reader have to think about how the parameters might be expressed in the code.

aineko/extras/connectors/rest_poller.py
ricrosales: Sure, but this seems better suited to showcase in the docs or in the docstring so it shows up in the docs.

aineko/extras/connectors/rest_poller.py
ricrosales: fixed

aineko/extras/connectors/rest_poller.py
craycrayfish: We should definitely move to the approach @Barabazs suggested. You would create a pydantic model in a python file (similar to `aineko/models/config_schema.py`), and import that into the node. When you receive `node_params`, you would do `RESTParam(params)` to automatically check the schema.

aineko/extras/connectors/rest_poller.py
ricrosales: they just have to be available in the environment. i'm intentionally not being prescriptive about how they get there. so a user could use a .env file or they could use source to add the environment variables or otherwise make those variables available to the environment.

Yes, the idea is that the user would reference those in the pipeline config using the {$) pattern.

I modified a bit, but how would you modify this documentation?
```
    Secrets can be injected (from environment) into the `url`, `headers`, and
    `data` fields by passing a string with the following format:
    `{$SECRET_NAME}`. For example, if you have an environment variable named 
    `SECRET_NAME`that contains the value `SECRET_VALUE`, you can inject it into 
    the url field by passing `https://example.com?secret={$SECRET_NAME}` as the 
    url. The connector will then replace `{$SECRET_NAME}` with `SECRET_VALUE` 
    before connecting to the REST endpoint.
```

note this is also documented here: https://docs.aineko.dev/dev/developer_guide/pipeline_configuration/#node_name

aineko/extras/connectors/rest_poller.py
craycrayfish: I agree with @Barabazs that we should make this more concise and put the details in the argument explanation itself. When I said default behavior I had a more generalized description in mind, like
`By default, the node will poll the endpoint every 5 seconds, and timeout after 10 seconds. It will retry forever if the connection fails. Any 20X response will be considered a success. No headers are added by default. `

aineko/extras/connectors/rest_poller.py
ricrosales: Yes. Since we sleep, it's cheap to restart the session.

aineko/extras/connectors/rest_poller.py
mdigi14: I think the documentation is okay. Just not great UX. We should return to this at some point and think of a cleaner solution.

aineko/extras/connectors/rest_poller.py
ricrosales: fair... but how would we rename this? APIPoller? GETPoller?

ricrosales: I'm changing this to HTTPPoller

aineko/extras/connectors/rest_poller.py
ricrosales: good idea

aineko/extras/connectors/rest_poller.py
giddyphysicist: That makes sense -- thank you. 

I think that's something we could revisit in the future if we settle on a standard recommended client library later.

aineko/extras/connectors/rest_poller.py
ricrosales: By construction, we reach this logic if and only if `time.time()` - `self.last_poll_time` < `self.rest_params.poll_interval` so thus the scenario you describe would never occur. If the API takes 2 seconds to respond, then on the next iteration through the loop, the code would flow into the polling logic.

mkdocs.yml
```diff
@@ -104,7 +104,8 @@ nav:
       - Extras:
           - extras/intro.md
           - extras/fastapi.md
-          - extras/websocket.md
+          - extras/websocket_client.md
```
Barabazs, line 107:
```suggestion
          - extras/websocket.md
```
The file is named websocket, not websocket_client

aineko/extras/connectors/http_poller.py
```diff
@@ -0,0 +1,257 @@
...
+    ]
+
+    @field_validator("url")
+    @classmethod
+    def supported_url(cls, url: str) -> str:
+        """Validates that the url is a valid HTTP or HTTPS URL."""
+        if not (url.startswith("https://") or url.startswith("http://")):
+            raise ValueError(
+                "Invalid url provided to HTTPS params. "
+                'Expected url to start with "https://". '
```
Barabazs: This still implies we only allow for https URL.

aineko/extras/connectors/http_poller.py
```diff
@@ -0,0 +1,257 @@
...
+            Defaults to 5.0.
+        max_retries (int): The maximum number of times to retry connecting to
+            the HTTP endpoint. Defaults to -1.
+        metadata (Optional[Dict[str, Any]]): A dictionary of metadata to attach
+            to outgoing messages. Defaults to None.
+        retry_sleep (float): The number of seconds to wait between retries.
+            Defaults to 5.0.
+        success_codes (List[int]): A list of HTTP status codes that indicate
+            success. Defaults to
+                [200, 201, 202, 203, 204, 205, 206, 207, 208, 226].
```
Barabazs: The type hints are not needed here. They will automatically be inferred on the docs.

aineko/extras/connectors/http_poller.py
```diff
@@ -0,0 +1,257 @@
...
+            dictionary, list of tuples, bytes, or file-like object to send in
+            the body of the HTTP request. Defaults to None.
+        json_ (Optional[Dict[str, Any]]): A JSON serializable Python object to
+            send in the body of the HTTP request. Defaults to None.
+        auth (Optional[Tuple[str, str]]): A tuple of username and password to
+            use for Basic HTTP authentication. Defaults to None.
+        poll_interval (float): The number of seconds to wait between polls.
+            Defaults to 5.0.
+        max_retries (int): The maximum number of times to retry connecting to
+            the HTTP endpoint. Defaults to -1.
```
Barabazs: I think you should explain what -1 does in this context.

mdigi14 (commented):
There are still a few references to REST in the code, and the PR title still references REST.

aineko/extras/connectors/http_poller.py
```diff
@@ -0,0 +1,257 @@
...
+                raw_message = response.text
+            except Exception as err:  # pylint: disable=broad-except
+                # If request fails, log the error and sleep
+                self.log(
+                    "Request failed. "
+                    f"Sleeping for {self.http_poller_params.poll_interval} "
+                    f"seconds. Error: {err}",
+                    level="error",
+                )
+                time.sleep(self.http_poller_params.poll_interval)
```
mdigi14: It's not clear to me why we use `poll_interval` here but `retry_sleep` in the exception handling below. 

If we're going to handle the cases differently then the documentation for `retry_sleep` should specify that it is only for parsing / writing to datasets.

In my mind, `The number of seconds to wait between retries.` implies that `retry_sleep` would be used whenever we need to retry something.

mdigi14: Also, if an endpoint is rate limiting the polls we might want to give the user more flexibility to wait longer than in the general case (and this would be tunable via the separate `retry_sleep` param)

aineko/extras/connectors/http_poller.py
```diff
@@ -0,0 +1,257 @@
...
+            )
+        self.output_dataset = output_datasets[0]
+
+        # Create a session
+        self.log(
+            f"Creating new session to endpoint {self.http_poller_params.url}."
+        )
+        self.session = requests.Session()
+
+    def _execute(self, params: Optional[Dict] = None) -> None:
```
mdigi14, line 166: Time permitting, I would break this function into smaller functions based on `polling` and `parse_and_write`. It's quite a bit longer than the ideal function length of ~40 lines. https://google.github.io/styleguide/pyguide.html#318-function-length

mkdocs.yml
ricrosales: fixed the name of the file instead

aineko/extras/connectors/http_poller.py
ricrosales: this was a mistake, i replaced poll_interval with retry_sleep here. im going to follow up with exponential backoff which should help in the rate limiting scenario you desribe.

aineko/extras/connectors/http_poller.py
ricrosales: fixed

ricrosales:
> There are still a few references to REST in the code, and the PR title still references REST.

I removed all references

Barabazs (commented):
To unblock.

Edit: seems like this doesn't reset the change request. Please ignore

aineko/extras/connectors/http_poller.py
```diff
@@ -0,0 +1,264 @@
...
+    This node is a wrapper around the
+    [requests](
+        https://docs.python-requests.org/en/master/
+        ){:target="_blank"} library.
+
+    Example usage in pipeline.yml:
+    ```yaml title="pipeline.yml"
+    pipeline:
+      nodes:
+        HTTPoller:
```
giddyphysicist: small typo (not crucial to operation): HTTPoller --> HTTPPoller

giddyphysicist (approved):
small comment above, but am approving. looks good!

aineko/extras/connectors/http_poller.py
```diff
@@ -0,0 +1,264 @@
...
+    Attributes:
+        timeout: The number of seconds to wait for the endpoint to
+            respond. Defaults to 10.
+        url: The URL to connect to.
+        headers: A dictionary of headers to send to the endpoint. Defaults
+            to None.
+        data: A dictionary of data to send to the endpoint. Defaults to None.
+        params: A dictionary, list of tuples, bytes, or file-like object to
+            send in the body of the request. Defaults to None.
+        json_: A JSON serializable Python object to send in the body of the
```
mdigi14, line 27: is this `json_` because json is a keyword?

aineko/extras/connectors/http_poller.py
```diff
@@ -0,0 +1,264 @@
...
+            # Reset the session
+            self.log(
+                "Creating new session to HTTP endpoint "
+                f"{self.http_poller_params.url}."
+            )
+            self.session = requests.Session()
+            return None
+
+    def parse_data(self, raw_data: str) -> Optional[Dict[str, Any]]:
+        """Parses raw endpoint response using JSON parser."""
```
mdigi14: We should include data on args / raises but I'm OK merging this PR for now given time constraints.

mdigi14: same with `poll_endpoint`

mdigi14 (approved):
Minor feedback but shouldn't block this imo

craycrayfish (approved):
Amazing!

aineko/extras/connectors/http_poller.py
ricrosales: yes

ricrosales: or not a keyword but it throws a linting error, because it can be misleading. in the context we are using it, i think it makes sense to use json so i chose to just add the _

PR #134, opened 2024-01-18
feat: support more python versions

Barabazs:
## Summary

This PR allows for using Aineko on different python version: 3.8, 3.9, 3.10 & 3.11

## Motivation

Make Aineko more accessible by being less restrictive for the version.

## Author Checklist

- [x] Status checks pass.

## Reviewer Checklist


craycrayfish:
The complete combination of checks should not be done for every commit - let's trigger the version checks when we're going to release a new version. A potential criteria could be checking the branch name to be `chore/bump-aineko-version` etc, but we would need to set some standard there. Alternatively, we could do some diff checking for the version defined in pyproject.toml.

Barabazs:
> The complete combination of checks should not be done for every commit - let's trigger the version checks when we're going to release a new version. A potential criteria could be checking the branch name to be `chore/bump-aineko-version` etc, but we would need to set some standard there. Alternatively, we could do some diff checking for the version defined in pyproject.toml.

Why would you want to do that? 

I'm in favor of doing it for every push because:
- The total time spent on testing is identical to before since the tests run in parallel.
- Merging incompatible code and cleaning up later is inefficient.

craycrayfish:
> > The complete combination of checks should not be done for every commit - let's trigger the version checks when we're going to release a new version. A potential criteria could be checking the branch name to be `chore/bump-aineko-version` etc, but we would need to set some standard there. Alternatively, we could do some diff checking for the version defined in pyproject.toml.
> 
> Why would you want to do that?
> 
> I'm in favor of doing it for every push because:
> 
> * The total time spent on testing is identical to before since the tests run in parallel.
> * Merging incompatible code and cleaning up later is inefficient.

This would increase the CircleCI cost by 5x for those jobs. Maybe not an issue yet since the total costs aren't very high. I'm fine with merging as is and then figuring out how to optimize later when it becomes a problem.

.circleci/config.yml
```diff
@@ -16,19 +16,42 @@ workflows:
...
+                - "3.9"
+                - "3.10"
+                - "3.11"
+          python_version: << matrix.python_version >>
           filters:
             branches:
               ignore: develop
           prj_dir: aineko
-          python_version: << pipeline.parameters.python_version >>
+          # python_version: << matrix.python_version >>
```
craycrayfish: Can we remove the commented lines?

craycrayfish (approved):

.circleci/config.yml
Barabazs: Addressed in 37d7433a8df88bc56f2ebfe921d5753191d1d321

PR #140, opened 2024-01-23
feat: replace message wrapper dict with Pydantic model & document message consumption

### LINKED ISSUE #138: Document the structure of messages
It's not explicit anywhere in the documentation what the expected message structure is. Per the `DatasetProducer` code, a message has the tags "timestamp", "dataset", "source_pipeline", "source_node", and "message". But this is not documented anywhere.

Barabazs:
## Summary

This PR replaces the dict which wraps around the produced message with a Pydantic model. This is in preparation of a future PR to allow for Pydantic models for the message itself.

I also added a paragraph to the docs about how to consume messages and the other attributes you can access.

Closes #138 
## Motivation

Using a python object instead of a dict makes it easier to access the attributes and document the use of it.

## Author Checklist

- [x] Status checks pass.

## Reviewer Checklist

- [x] Title is accurate and formatted appropriately.
- [x] Description motivates each change.
- [x] Docs are updated if necessary.
- [x] Code is pythonic and consistent with the rest of the codebase.
- [x] The code is consistent with the [style guide](https://google.github.io/styleguide/pyguide.html).


mdigi14:
@Barabazs Can you add an example of what the messages look like before and after this change? I'm having a tough time understanding how this is represented at a lower level and how it would affect some of the monitoring tools we've built.

If I'm understanding correctly we have JSON in both cases, but in the updated code the JSON represents the `MessageData` class instead of just being a JSON representation of a Python `dict`

mdigi14:
When you say 

"Using a python object instead of a dict makes it easier to access the attributes and document the use of it."

do you just mean that the code is cleaner / easier to follow when using class fields as opposed to dict keys that aren't explicitly defined in a model?

docs/developer_guide/node_implementation.md
```diff
@@ -167,6 +167,17 @@ In cases where you might require more low-level control over consumption pattern
 
 The timeout argument in these methods signify the duration in which the method has to return a message otherwise it will re-poll for a new one.
 
+#### Consuming Messages
```
mdigi14, line 170: This is good to document. Nice work

Barabazs:
> @Barabazs Can you add an example of what the messages look like before and after this change? I'm having a tough time understanding how this is represented at a lower level and how it would affect some of the monitoring tools we've built.
> 
> If I'm understanding correctly we have JSON in both cases, but in the updated code the JSON represents the `MessageData` class instead of just being a JSON representation of a Python `dict`

Nothing is changed in regards to how the message is produced/consumed. It's still dict -> json str -> bytes.

What actually changes is the returned value from the various consume methods. You used to get a dict and you would access fields like: 
```python
x = {
    "timestamp": "2024-xxx",
    "dataset": "dataset_name",
    "source_pipeline": "source_pipeline_name",
    "source_node": "source_node_name",
    "message": {"foo": "bar"},
 }

x["timestamp"]
```

And now you would get a python object where you can access the attributes directly:
```python
x = MessageData(
    timestamp="2024-xxx",
    dataset="dataset_name",
    source_pipeline="source_pipeline_name",
    source_node="source_node_name",
    message={"foo": "bar"}
)

x.timestamp
```

> When you say
> 
> "Using a python object instead of a dict makes it easier to access the attributes and document the use of it."
> 
> do you just mean that the code is cleaner / easier to follow when using class fields as opposed to dict keys that aren't explicitly defined in a model?

Yes, exactly.

.circleci/config.yml
mdigi14, line 1: Why is this being updated? I'm not sure if this is related to the core changes

Barabazs: See b39ea7d6fe42d68eb06c2bc6da2025d9b62fbe09 and 1e846191d0bdd16946c672076b57dc3dd56e4021.

In short:
* running pytest without verbosity make the entire CI run useless if you can't see why it failed
* For poetry it's also useless to run it as --quiet, because you would get a failure and no reason

aineko/core/dataset.py
```diff
@@ -381,30 +396,30 @@ def _delivery_report(err: Any, message: Message) -> None:
...
+            message=message,
+            source_node=self.source_node,
+            source_pipeline=self.source_pipeline,
+        ).model_dump(mode="json")
+
         self.producer.poll(0)
 
         key_bytes = str(key).encode("utf-8") if key is not None else None
 
         self.producer.produce(
```
mdigi14, line 419: My main concern here was ensuring this output looked consistent with what we've built so far and it looks good 👍 

Example: 

```
wrapped_message = MessageData(
    dataset="logging",
    message="Hello World!",
    source_node="logging_node",
    source_pipeline="logging_pipeline",
)
```

produces

`b'{"message": "Hello World!", "timestamp": "2024-01-25 11:22:25.463134", "dataset": "logging", "source_pipeline": "logging_pipeline", "source_node": "logging_node"}'
`

mdigi14: So there shouldn't be any issues with compatibility

.circleci/config.yml
mdigi14: Is this related to this PR though? I'm fine pushing this through but it seems like a separate fix unless changes in this PR necessitate the change

aineko/models/base.py
```diff
@@ -0,0 +1,29 @@
...
+"""Base models for Aineko."""
+import datetime
+from typing import Union
+
+from pydantic import BaseModel, Field
+
+from aineko.config import AINEKO_CONFIG
+
+
+class MessageData(BaseModel):
```
mdigi14, line 12: I'm not sure `MessageData` is a good name for this class. Especially since the Kafka implementation we use has a class named `Message`. Why not use `WrappedMessage`? We often use `wrapped_message = MessageData` in the code

mdigi14 (approved):
Looks good to merge, but left some comments related to scope of PR and naming conventions. 

Also confirmed that the low-level representation of the data seems consistent.

.circleci/config.yml
Barabazs: It's not, but last time I was asked to merge small PRs together 🤷‍♂️

aineko/models/base.py
Barabazs: I initially chose WrappedMessage, but I felt like the name was not self-explanatory enough for users. But maybe MessageData is worse.

.circleci/config.yml
Barabazs: I reverted the changes and opened a separate PR.

aineko/models/base.py
Barabazs: Posted in Slack for feedback: https://aineko-dev.slack.com/archives/C05U15ATA2V/p1706450232466239

Barabazs:
Closing this in favor of the upcoming changes to datasets

PR #150, opened 2024-02-02
feat: implement abstract dataset and refactor Kafka dataset

giddyphysicist:
## Summary

Create new AbstractDataset base class so that new dataset types can inherit from this general class.

Refactor the Kafka dataset to inherit from AbstractDataset

Update node.py and runner.py to handle the more general syntax for the Kafka Dataset.

Update tests

## Motivation

To have a standard way of adding more general dataset types to support types such as batch, streaming, or in-memory datasets.

## Author Checklist

- [x] Status checks pass.

## Reviewer Checklist

- [x] Title is accurate and formatted appropriately.
- [x] No unnecessary changes are introduced.
- [x] Description motivates each change.
- [x] Docs are updated if necessary.
- [x] Code is pythonic and consistent with the rest of the codebase.
- [x] The code is consistent with the [style guide](https://google.github.io/styleguide/pyguide.html).


Barabazs:
Is there a specific reason to move the dataset implementations outside of `aineko.core.dataset`?

What if we went from 
```
├── __init__.py
├── __main__.py
├── cli
│   ├── ...
├── config.py
├── core
│   ├── __init__.py
│   ├── config_loader.py
│   ├── dataset.py (old code?)
│   ├── deploy_config_loader.py
│   ├── node.py
│   ├── node_manager.py
│   └── runner.py
├── datasets
│   ├── __init__.py
│   ├── core.py
│   └── kafka.py
├── ...
...
```

to:
```
├── __init__.py
├── __main__.py
├── cli
│   ├── ...
├── config.py
├── core
│   ├── __init__.py
│   ├── config_loader.py
│   ├── dataset.py (AbstractDataset)
│   ├── deploy_config_loader.py
│   ├── node.py
│   ├── node_manager.py
│   └── runner.py
├── datasets (out of the box Aineko dataset implementations)
│   ├── __init__.py
│   └── kafka.py
├── ...
...
```

giddyphysicist:
> Is there a specific reason to move the dataset implementations outside of `aineko.core.dataset`?
> 
> What if we went from
> 
> ```
> ├── __init__.py
> ├── __main__.py
> ├── cli
> │   ├── ...
> ├── config.py
> ├── core
> │   ├── __init__.py
> │   ├── config_loader.py
> │   ├── dataset.py (old code?)
> │   ├── deploy_config_loader.py
> │   ├── node.py
> │   ├── node_manager.py
> │   └── runner.py
> ├── datasets
> │   ├── __init__.py
> │   ├── core.py
> │   └── kafka.py
> ├── ...
> ...
> ```
> 
> to:
> 
> ```
> ├── __init__.py
> ├── __main__.py
> ├── cli
> │   ├── ...
> ├── config.py
> ├── core
> │   ├── __init__.py
> │   ├── config_loader.py
> │   ├── dataset.py (AbstractDataset)
> │   ├── deploy_config_loader.py
> │   ├── node.py
> │   ├── node_manager.py
> │   └── runner.py
> ├── datasets (out of the box Aineko dataset implementations)
> │   ├── __init__.py
> │   └── kafka.py
> ├── ...
> ...
> ```

Yes I think your suggestion makes more sense -- having the base classes (for both nodes and datasets) under the core directory. Having a separate datasets folder would then make it easier to separate as plugins as well.

aineko/core/dataset.py
```diff
@@ -1,525 +1,276 @@
...
+
+    @abc.abstractmethod
+    def _initialize(self, *args: T, **kwargs: T) -> Any:
+        """Subclass implementation to initialize the dataset query layer."""
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def _describe(self, *args: T, **kwargs: T) -> str:
+        """Subclass implementation to describe the dataset metadata."""
+        return f"Dataset name: {self.name}"
```
Barabazs: What's the point of this method?

aineko/core/dataset.py
```diff
@@ -1,525 +1,276 @@
...
-        """Initialize the producer."""
-        self.dataset_name = dataset_name
-        self.node_name = node_name
-        self.values = []  # type: ignore
+    name: str
 
-    def produce(self, message: Any) -> None:
-        """Stores message in self.values.
+    @classmethod
+    def from_config(cls: Type[A], name: str, config: Dict[str, Any]) -> A:
```
Barabazs:
```suggestion
    def from_config(cls, name: str, config: Dict[str, Any]) -> AbstractDataset:
```
cls doesn't need to be type hinted and we know that from_config will return AbstractDataset

aineko/core/dataset.py
```diff
@@ -1,525 +1,276 @@
...
+        """Subclass implementation to initialize the dataset query layer."""
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def _describe(self, *args: T, **kwargs: T) -> str:
+        """Subclass implementation to describe the dataset metadata."""
+        return f"Dataset name: {self.name}"
+
+    @abc.abstractmethod
+    def _exists(self, *args: T, **kwargs: T) -> bool:
```
Barabazs: Do we actually expect to pass anything to this method?

aineko/core/dataset.py
```diff
@@ -1,525 +1,276 @@
...
+        """Subclass implementation to delete the dataset."""
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def _initialize(self, *args: T, **kwargs: T) -> Any:
+        """Subclass implementation to initialize the dataset query layer."""
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def _describe(self, *args: T, **kwargs: T) -> str:
```
Barabazs: Do we actually expect to pass anything to this method?

aineko/core/dataset.py
giddyphysicist: This is supposed to be a common place to return metadata about a specific dataset. By default, it just returns the name, but particular implementations can add data to this. It may not be strictly necessary (and if it is going to be a string, we could just make it the `__str__` or `__repr__` method)

aineko/core/dataset.py
giddyphysicist: I get mypy errors when I try implementing it like this (remove Type[A] and change return type to AbstractDataset):
```Undefined variable 'AbstractDataset' (undefined-variable)```

If I try to remove Type[A] and replace the return type as A, I get a different mypy error:
```error: A function returning TypeVar should receive at least one argument containing the same TypeVar  [type-var]```

I think this has to do with from_config returning a subclass of AbstractDataset

File: aineko/core/dataset.py (hunk 3 above)
Comment (giddyphysicist) [reply aineko/core/dataset.py]: Not for the immediate dataset implementations we have in mind. But I wanted to allow flexibility for other implementations if it is needed. However, if you think it is unnecessary, we could always remove it now and add it back later if it is needed.

File: aineko/core/dataset.py (hunk 4 above)
Comment (giddyphysicist) [reply aineko/core/dataset.py]: Not for our current dataset implementations, but it's possible users who make a different implementation will want something more flexible (like filtering), depending on the structure or how extensive the metadata is.

File: aineko/core/dataset.py (hunk 2 above)
Comment (giddyphysicist) [reply aineko/core/dataset.py]: Oh I see, if I put quotes around AbstractDataset for the return type (and remove Type[A]), it doesn't throw mypy errors. 
I do think what you propose is simpler, but also might have slightly less type safety.

**Comment by giddyphysicist** [comment]
I pushed an update after merging with develop. Note, since the extras were moved to the plugins repo, we will need to update the tests in the plugins repo whenever this PR is merged (compare with tests at commit [6de13ab](https://github.com/aineko-dev/aineko/pull/150/commits/6de13ab7ae4a222edeef9f854859f81aea2f88f0) for reference for how to update the plugins tests)

File: aineko/datasets/kafka.py [hunk 5]
```diff
@@ -0,0 +1,839 @@
...
+        DEFAULT_KAFKA_CONFIG.CONSUMER_CONFIG[
+            "bootstrap.servers"
+        ] = self.location
+        DEFAULT_KAFKA_CONFIG.PRODUCER_CONFIG[
+            "bootstrap.servers"
+        ] = self.location
+
+
+# pylint: enable=too-few-public-methods
+# pylint: disable=unused-argument
```
Barabazs:
```suggestion
# pylint: enable=too-few-public-methods
# pylint: disable=unused-argument
```
The first disable rule is obsolete afaik and the unused-argument rule should be handled in the method itself.  (See paragraph about unused arguments here: https://google.github.io/styleguide/pyguide.html#214-decision)

So in the `read` method you would add something like
`del timeout  # unused because this is a fake dataset` to the first line after the docstring.

aineko/datasets/kafka.py
Barabazs, line 1: I notice a lot of inconsistent use of Consumer/Producer vs Input/Output. (both here as in other files)

aineko/models/config_schema.py
```diff
@@ -16,6 +16,7 @@ class Dataset(BaseModel):
             """Dataset model."""
 
             type: str
+            location: Optional[str]
```
Barabazs:
```suggestion
            location: Optional[str] = None
```

aineko/datasets/kafka.py
```diff
@@ -0,0 +1,839 @@
...
+        self.node_name = node_name
+        self.values = []  # type: ignore
+
+    def write(self, message: Any) -> None:
+        """Stores message in self.values.
+
+        Args:
+            message: message to produce.
+        """
+        self.values.append(message)
```
Barabazs: This isn't consistent with the new dataset implementation. This should either be merged into one "FakeKafka" or (preferably) be handled in the main Kafka class.

mkdocs.yml
```diff
@@ -102,12 +102,12 @@ nav:
...
-          - api_reference/dataset_producer.md
-          - api_reference/config.md
-          - api_reference/config_loader.md
-          - api_reference/runner.md
+          - 'api_reference/abstract_node.md'
+          - 'api_reference/abstract_dataset.md'
+          - 'api_reference/kafka_dataset.md'
+          - 'api_reference/config.md'
+          - 'api_reference/config_loader.md'
+          - 'api_reference/runner.md'
```
Barabazs: The quotation marks are not needed here. This adds extra diffs to the git history and is also inconsistent with the rest of the file.

tests/conftest.py
```diff
@@ -74,8 +74,10 @@ class DummyNode(AbstractNode):
 
         def _execute(self, params: Optional[dict] = None) -> Optional[bool]:
             """Consumes message from input and outputs it to output."""
-            msg = self.consumers["input"].consume(how="next", timeout=0)
-            self.producers["output"].produce(msg)
+            print(self.inputs)
+            print(self.outputs)
```
Barabazs:
```suggestion
```
I suppose this was used while developing, but this doesn't add value to the tests.

tests/integration/test_runner_with_kafka.py
```diff
@@ -123,21 +136,35 @@ def test_write_read_to_kafka(start_service, subtests):
...
+                    ),
+                }
+            )
+            dataset.initialize(
+                create_consumer=True, connection_params=consumer_params
+            )
+            count_messages = dataset.consume_all(end_message="END")
+            print("count_messages are...")
+            print(count_messages)
+            print("...")
```
Barabazs:
```suggestion
```

tests/integration/test_runner_with_kafka.py
```diff
@@ -123,21 +136,35 @@ def test_write_read_to_kafka(start_service, subtests):
...
+            # assert count_messages[0]["source_pipeline"] == "integration_test_read"
```
Barabazs: Why is this commented out?

aineko/datasets/__init__.py
Barabazs, line 1: Have you considered importing the Kafka dataset here, so we can use it as aineko.datasets.Kafka ?

aineko/__init__.py
```diff
@@ -9,12 +9,7 @@
 import logging
 
 from aineko.core.config_loader import ConfigLoader
-from aineko.core.dataset import (
-    DatasetConsumer,
-    DatasetProducer,
-    FakeDatasetConsumer,
-    FakeDatasetProducer,
-)
+from aineko.core.dataset import AbstractDataset
```
Barabazs: I'm not sure why we would want to expose this at the root level of the package. A regular user has no use for this and an advanced user that wants to implement a custom dataset will know where to look by reading the docs.

aineko/datasets/kafka.py
```diff
@@ -0,0 +1,839 @@
...
+        ```python
+        self._create(topic_params=TopicParams(...))
+        ```
+    """
+
+    dataset_prefix: Optional[str] = None
+    dataset_config: Optional[Dict[str, Any]] = {}
+
+
+class Kafka(AbstractDataset):
```
Barabazs: I think we should rename this to `KafkaDataset`. Kafka is too ambiguous and generic, whereas KafkaDataset is very clear and specific to what the class is about.

tests/core/test_dataset.py
Barabazs, line 1: The file structure of the tests should match the actual file structure as much as possible. This should live under `/tests/datasets/test_kafka.py`.

Barabazs, line 1: The new KafkaDataset implementation does not have enough tests. Passing the existing tests is not enough when we add/modify that many features.

tests/conf/integration_test_empty_dataset.yml
```diff
@@ -16,6 +16,8 @@ pipeline:
 
   datasets:
     messages:
-      type: kafka_stream
+      type: aineko.datasets.kafka.Kafka
+      location: localhost:9092
```
Barabazs, line 20: We should have a few tests where location is missing, location is missing and not in ENV vars, etc.

aineko/core/dataset.py
```diff
@@ -1,529 +1,278 @@
...
+
+    def read(self, *args: T, **kwargs: T) -> Any:
+        """Read the dataset via the query layer."""
+        try:
+            return self._read(*args, **kwargs)
+        except DatasetError:
+            raise
+        except Exception as e:
+            message = f"Failed to read dataset {self.name}."
+            raise DatasetError(message) from e
```
Barabazs: All methods which raise an exception should mention it in the docstring.

aineko/core/dataset.py
Barabazs, line 1: What's the rationale for implementing the public methods as wrappers for the abstract internal methods? Why not define the required public methods as abstract methods instead?

aineko/core/node.py
```diff
@@ -118,36 +126,66 @@ def setup_datasets(
             prefix: prefix for topic name (`<prefix>.<dataset_name>`)
             has_pipeline_prefix: whether the dataset name has pipeline name
                 prefix
+
+        TODO: update docstring
```
Barabazs: TODO?

aineko/core/node.py
```diff
@@ -118,36 +126,66 @@ def setup_datasets(
...
-                    has_pipeline_prefix=has_pipeline_prefix,
+                dataset_name: AbstractDataset.from_config(
+                    name=dataset_name, config=datasets.get(dataset_name, {})
                 )
                 for dataset_name in outputs
             }
         )
+        # topics have already been created from runner.
+
+        # now create the producers and consumers connections:
```
Barabazs:
```suggestion
```

aineko/core/node.py
```diff
@@ -208,9 +248,10 @@ def log(self, message: str, level: str = "info") -> None:
                 f"{', '.join(self.log_levels)}"
             )
         out_msg = {"log": message, "level": level}
-        self.producers[DEFAULT_KAFKA_CONFIG.get("LOGGING_DATASET")].produce(
-            out_msg
-        )
+        # self.producers[DEFAULT_KAFKA_CONFIG.get("LOGGING_DATASET")].produce(
+        #     out_msg
+        # )
```
Barabazs: To be deleted?

aineko/core/node.py
```diff
@@ -358,15 +399,20 @@ def run_test_yield(
                     continue
 
             # End loop if all consumers are empty
-            if self.consumers and all(
-                consumer.empty for consumer in self.consumers.values()
+            # if self.consumers and all(
+            #     consumer.empty for consumer in self.consumers.values()
+            # ):
+            #     run_loop = False
```
Barabazs: ?

aineko/core/runner.py
```diff
@@ -139,88 +139,65 @@ def prepare_datasets(
...
+            "location": "localhost:9092",
+            "type": "aineko.datasets.kafka.Kafka",
+        }
+        logging_dataset_name = DEFAULT_KAFKA_CONFIG.get("LOGGING_DATASET")
+        logger.info(
+            "Creating dataset: %s: %s", logging_dataset_name, logging_config
+        )
+        logging_dataset: AbstractDataset = AbstractDataset.from_config(
+            logging_dataset_name, logging_config
+        )
```
Barabazs: The type hints don't seem correct. The datasets are not AbstractDataset in this context.

aineko/core/node.py
```diff
@@ -65,8 +66,15 @@ class AbstractNode(ABC):
     The _execute method is where the node logic is implemented by the user.
 
     Attributes:
+        name (str): name of the node
+        pipeline_name (str): name of the pipeline
+        params (dict): dict of parameters to be used by the node
         consumers (dict): dict of DatasetConsumer objects for inputs to node
         producers (dict): dict of DatasetProducer objects for outputs of node
```
Barabazs: Are these still attributes?

Barabazs (changes requested):

Barabazs:
Please mark the PR as "ready for review" when I should revisit. Thanks

tests/integration/test_runner_with_kafka.py
giddyphysicist: i added it back in. originally, i wasn't going to implement since we're updating the message structure to remove metadata. but i think that's something that should wait until the dataset validation PR

aineko/datasets/kafka.py
giddyphysicist: I had originally called it KafkaDataset, but I think @craycrayfish preferred to call it something simpler. I'm fine either way, but would lean towards KafkaDataset

tests/core/test_dataset.py
giddyphysicist: updated

aineko/core/node.py
giddyphysicist: forgot to remove this, thanks

aineko/core/node.py
giddyphysicist: deleted

aineko/core/node.py
giddyphysicist: no, i removed it.

aineko/core/node.py
giddyphysicist: old code, i removed

tests/integration/test_runner_with_kafka.py
Barabazs: Yes, we shouldn't modify anything that's out of scope for this feature.

aineko/datasets/kafka.py
Barabazs: Let's go with KafkaDataset.

aineko/datasets/kafka.py
giddyphysicist: updated with a combined class FakeKafka

aineko/datasets/kafka.py
giddyphysicist: I made a pass through to update this, thanks

aineko/datasets/__init__.py
giddyphysicist: I hadn't considered it. Do you think this would have a large benefit?

aineko/__init__.py
giddyphysicist: I was mainly thinking about the symmetry between AbstractNode and AbstractDataset. Also, we access the AbstractDataset.from_config method from the pipeline Runner as a generalized way of loading in subclasses of AbstractDataset. I'm ok with removing it from here, but it seems to balance with AbstractNode

aineko/datasets/kafka.py
giddyphysicist: Sounds good. I updated it to KafkaDataset

tests/conf/integration_test_empty_dataset.yml
giddyphysicist: Good call, I added a few tests in.

aineko/core/dataset.py
giddyphysicist: Added.

aineko/core/dataset.py
giddyphysicist: I think this was mainly to add generic error handling (and potentially more future features like logging or telemetry) as a standard wrapper. This also similarly follows the format for `execute` wrapping user-implemented  `_execute` method for AbstractNode subclasses

craycrayfish:
> Is there a specific reason to move the dataset implementations outside of `aineko.core.dataset`?
> 
> What if we went from
> 
> ```
> ├── __init__.py
> ├── __main__.py
> ├── cli
> │   ├── ...
> ├── config.py
> ├── core
> │   ├── __init__.py
> │   ├── config_loader.py
> │   ├── dataset.py (old code?)
> │   ├── deploy_config_loader.py
> │   ├── node.py
> │   ├── node_manager.py
> │   └── runner.py
> ├── datasets
> │   ├── __init__.py
> │   ├── core.py
> │   └── kafka.py
> ├── ...
> ...
> ```
> 
> to:
> 
> ```
> ├── __init__.py
> ├── __main__.py
> ├── cli
> │   ├── ...
> ├── config.py
> ├── core
> │   ├── __init__.py
> │   ├── config_loader.py
> │   ├── dataset.py (AbstractDataset)
> │   ├── deploy_config_loader.py
> │   ├── node.py
> │   ├── node_manager.py
> │   └── runner.py
> ├── datasets (out of the box Aineko dataset implementations)
> │   ├── __init__.py
> │   └── kafka.py
> ├── ...
> ...
> ```

The main reason I proposed this was because I imagine Aineko to have several decoupled components in the future - runner, datasets, config loader, deployment config. And they wold have their own directories eventually. So it makes sense to start decoupling them

aineko/core/runner.py
giddyphysicist: This one is a bit tricky. They are all technically subclasses of AbstractDataset, which can by implied by the syntax. They aren't all necessarily a single type of subclass (like KafkaDataset) since we want the runner to load the most general case of datasets. I couldn't find a recommended way of handling this, other than referencing the generic superclass of AbstractDataset

tests/core/test_dataset.py
giddyphysicist: Thanks. I've added a few tests, and I think the refactored original tests cover a good bit of the functionality. Are there any particular places you recommend more testing?

aineko/core/dataset.py
Barabazs: I don't think this pattern is useful or worth the trade offs. The most important downside (imo) is that we're hiding the useful information from the user. They don't know what a method does, which arguments to pass and what it's returning:

image

vs

image

I would suggest removing the wrapper methods and making the public methods abstract methods.

aineko/__init__.py
Barabazs: > I was mainly thinking about the symmetry between AbstractNode and AbstractDataset.

I'm not sure I understand what you mean by this.

The AbstractNode is used by the user when creating their own node. And every user needs to create a node, so it makes sense to expose it at the root level. But that's not the case for the AbstractDataset.

> Also, we access the AbstractDataset.from_config method from the pipeline Runner

True, but that doesn't require us to expose it at this level.

aineko/datasets/__init__.py
Barabazs: Imo it could, but maybe it's worth waiting until we have more dataset types and reassess. I was mainly thinking about how it could look like for the pipeline configs:
```diff
- type: aineko.datasets.kafka.KafkaDataset
+ type: aineko.datasets.KafkaDataset
```

But the aineko-plugins won't be able to support that since the first 2 names are namespace packages... (example `aineko_plugins.datasets.postgres.PostgresDataset`)

docs/developer_guide/node_implementation.md
```diff
@@ -63,20 +63,20 @@ class MySumNode(AbstractNode):
...
 ```
 
 A node will only terminate when the entire pipeline goes down or when the [poison pill](#poison-pill) is activated. 
 
 
-### Producers & consumers
+### Inputs & Outputs
 
-Node classes inherit attributes named `self.producers` and `self.consumers` that are each a dictionary, with keys being the dataset name and values being `DatasetProducer` and `DatasetConsumer` objects respectively. These objects allow you to produce/consume data to/from a dataset.
+Node classes inherit attributes named `self.inputs` and `self.outputs` that are each a dictionary, with keys being the dataset name and values being subclasses of `AbstractDataset`. These objects allow you to read/write data to/from a dataset.
```
Barabazs:
```suggestion
Node classes inherit attributes named `self.inputs` and `self.outputs` that are each a dictionary, with keys being the dataset name and values being subclasses of `AbstractDataset`. These objects allow you to read/write data from/to a dataset.
```
You switched the order, so it should also be updated int rest of the sentence.

docs/developer_guide/node_implementation.md
```diff
@@ -91,78 +91,78 @@ class MySumNode(AbstractNode):
...
         )
         self.state = int(msg["message"]) + int(params["increment"])
-        self.producers["test_sum"].produce(self.state)
+        self.outputs["test_sum"].write(self.state)
 ```
 
 
-!!! warning "Producers and Consumers must be included in the pipeline configuration"
+!!! warning "Inputs and Outputs must be included in the pipeline configuration"
     They must be defined in the `outputs` and `inputs` list respectively to be available to the node. If a dataset is not available in a Node's catalog, a `KeyError` will be raised.
```
Barabazs:
```suggestion
!!! warning "Inputs and Outputs must be included in the pipeline configuration"
    They must be defined in the `inputs` and `outputs` list respectively to be available to the node. If a dataset is not available in a Node's catalog, a `KeyError` will be raised.
```

docs/index.md
```diff
@@ -99,8 +99,8 @@ Supercharge your development by using pre-built nodes directly in your pipeline.
 ## API Documentation
 
 - [`AbstractNode`](./api_reference/abstract_node.md)
-- [`DatasetConsumer`](./api_reference/dataset_consumer.md)
-- [`DatasetProducer`](./api_reference/dataset_producer.md)
+- [`AbstractDataset`](./api_reference/abstract_dataset.md)
+- [`Kafka`](./api_reference/kafka_dataset.md)
```
Barabazs:
```suggestion
- [`KafkaDataset`](./api_reference/kafka_dataset.md)
```

tests/integration/test_runner_with_kafka.py
```diff
@@ -102,19 +102,30 @@ def test_write_read_to_kafka(start_service, subtests):
         try:
             runner.run()
         except ray.exceptions.RayActorError:
```
Barabazs:
```suggestion
        except ray.exceptions.RayActorError:
            # This is expected because we activated the poison pill
            pass
        ....
```
You should keep the same structure when doing this try-except block with a PoisonPill activation. (see comments in a previous PR: https://github.com/aineko-dev/aineko/pull/112#pullrequestreview-1794510059)

docs/index.md
Barabazs, line 1: Commenting here since I can't comment on an unmodified file:
The `docs/examples/aineko_dream.md` still uses the old notation of `self.consumers` etc

Barabazs (changes requested):
Thank you for addressing the feedback. There are still some issues to resolve, but we're getting closer to merging.

Barabazs (commented):
I think the kafka related parts are still too tightly intertwined with the rest of the codebase. (examples are node or runner) We should definitely revisit this in a new PR.

Barabazs:
@giddyphysicist can you respond to the feedback conversations once it's addressed instead of resolving it please? Otherwise it becomes challenging to follow up properly if everything is marked as resolved.

aineko/core/dataset.py
Barabazs: Let's remove it and add a `__str__` method instead. (just before the  `from_config` method.
I suggest:
```python
    def __str__(self) -> str:
        """Return the string representation of the dataset."""
        return f"{self.__class__.__name__}({self.name})"
```

<details>
<summary>
Reasoning
 </summary>

* I don't see any benefit in having it and making this part of the abstract dataset.
* It's too broad and unconstrained, which results in a different implementation for each dataset. For example:
  * kafka dataset: https://github.com/aineko-dev/aineko/blob/4544e16ba310661433d57a8d5d08a38a0f78a922/aineko/datasets/kafka.py#L351-L365
  * postgres dataset: https://github.com/aineko-dev/aineko/blob/e59bc2f46a406b1a1c830985206a2b45ecd05b0f/aineko/extras/datasets/postgres.py#L142-L143
* It's always easier to add a method later than to deprecate and remove one.

  </details>

giddyphysicist: this makes more sense. updated this.

aineko/__init__.py
giddyphysicist: That sounds fine. I updated.

aineko/core/dataset.py
giddyphysicist: Yes, I think this makes more sense. Especially with the method details. I updated.

docs/developer_guide/node_implementation.md
giddyphysicist: thanks, updated

docs/developer_guide/node_implementation.md
giddyphysicist: updated

docs/index.md
giddyphysicist: thanks, updated this

tests/integration/test_runner_with_kafka.py
giddyphysicist: Yes, that's more consistent. I updated.

docs/index.md
giddyphysicist: updated.

aineko/core/runner.py
Barabazs: Thanks, I see how that makes sense.

tests/core/test_dataset.py
Barabazs: Thanks! It seems like all the methods and features are tested (some indirectly). 🚀

Barabazs (approved):
This is a huge step forwards! Thank you  🚀 🔥

aineko/core/dataset.py
```diff
@@ -1,529 +1,213 @@
...
+        config={
+            "type": "aineko.datasets.mydataset.MyDataset",
+            "location": "foo",
+            "params": {
+                "param_1": "bar"
+            }
+        }
+    )
+    ```
     """
```
Barabazs: I just noticed that the docstring still mentions the method names with an underscore. This should be updated before merging please.

Barabazs (changes requested):

aineko/core/dataset.py
Barabazs: Updated it in  dc6552b98fc52a4cff6c3fcb05e7b8cab24c7210

Barabazs (approved):

aineko/core/node.py
```diff
@@ -120,35 +122,61 @@ def setup_datasets(
...
+                        "prefix": prefix,
+                        "has_pipeline_prefix": has_pipeline_prefix,
+                        "consumer_config": DEFAULT_KAFKA_CONFIG.get(
+                            "CONSUMER_CONFIG"
+                        ),
+                    }
+                )
+                self.inputs[dataset_name].initialize(
+                    create_consumer=True, connection_params=consumer_params
+                )
```
craycrayfish: This needs to be a separate function so we can initialize different dataset types here (and in the runner). This is fine if it will be refactored in a subsequent PR

aineko/datasets/kafka.py
```diff
@@ -0,0 +1,823 @@
...
+    """
+
+    dataset_name: str
+    node_name: str
+    pipeline_name: str
+    prefix: Optional[str] = None
+    has_pipeline_prefix: bool
+    producer_config: Dict[str, Any] = DEFAULT_KAFKA_CONFIG.get(
+        "PRODUCER_CONFIG"
+    )
```
craycrayfish: Can create a generic KafkaParams class which both of these inherit so we don't need to repeat the params

aineko/datasets/kafka.py
```diff
@@ -0,0 +1,823 @@
...
+        """
+        return self._create_topic(
+            dataset_name=self.name, topic_params=topic_params
+        )
+
+    def initialize(
+        self,
+        connection_params: Union[ConsumerParams, ProducerParams],
+        create_consumer: bool = False,
+        create_producer: bool = False,
```
craycrayfish: why not set the arg as create: Literal["consumer", "producer"] so that the fact that you can choose one is implicit

craycrayfish:
Amazing work! Thank you for the push!n Added a couple of non-blocking comments

craycrayfish (approved):

aineko/datasets/kafka.py
Barabazs: Maybe reader/writer to be consistent with the terms were using now?

giddyphysicist: Yes, this seems better. I updated.

aineko/datasets/kafka.py
giddyphysicist: That seems cleaner, thanks.

PR #159, opened 2024-02-16
fix: use -1 to block until the next message is available

### LINKED ISSUE #160: Unexpected behavior  - .next() is not blocking
Hey team - I'm taking a look at the new implementation of datasets and also the former (0.3.2) and realizing I may have introduced some unexpected behavior in the original design.

I'm looking at code
here (develop_: <https://github.com/aineko-dev/aineko/blob/develop/aineko/datasets/kafka.py#L423>
and here (0.3.2): <https://github.com/aineko-dev/aineko/blob/75bde7811ff8018e38202330927196c2476c70b2/aineko/core/dataset.py#L207>

I believe we intended the polling to lock in the default case when we use `.next()` on the consumer. However, looking at the confluent-kafka-python docs (<https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#pythonclient-consumer>) I'm seeing that locking forever only happens when you use `timeout=-1` (this is the default for confluent-kafka-python). Since we are passing in `timeout=None` by default, we are not actually locking. The `_consume_message` (<https://github.com/aineko-dev/aineko/blob/75bde7811ff8018e38202330927196c2476c70b2/aineko/core/dataset.py#L226|here>) method hides this since it hops into a while True loop.

We should fix this because this seems to be causing more resource consumption than necessary, I noticed this by inspecting Node consumption in Aineko Cloud and seeing that a Node that uses the `.next()` method was spinning on the CPU even though there are no messages to process.

Should we patch 0.3.2? Or what's the plan for releasing the new version of datasets?

[Slack Message](https://aineko-dev.slack.com/archives/C05U15ATA2V/p1708036808241439?thread_ts=1708036808.241439&cid=C05U15ATA2V)

Barabazs:
## Summary

This PR fixes the locking behavior of the KafkaDataset `.next()` method, as reported by @ricrosales on [Slack](https://aineko-dev.slack.com/archives/C05U15ATA2V/p1708036808241439).

Closes #160

## Motivation

Fix unintended behavior.

## Author Checklist

- [x] Status checks pass.

## Reviewer Checklist

### DIFF
--- a/aineko/datasets/kafka.py
+++ b/aineko/datasets/kafka.py
@@ -481,7 +481,7 @@ class KafkaDataset(AbstractDataset):
         Returns:
             message from the dataset
         """
-        return self._consume_message(how="next")
+        return self._consume_message(how="next", timeout=-1)
 
     def last(self, timeout: int = 1) -> Dict:
         """Consumes the last message from the dataset.


craycrayfish (dismissed):

craycrayfish:
I'm wondering if the `self.consume` should default to -1 rather than 0

Barabazs:
> I'm wondering if the `self.consume` should default to -1 rather than 0

@craycrayfish 
Don't you mean `_consume_message`? Because the docstring for _consume_message mentions that the method is supposed to be blocking. But it's not since timeout=0

ricrosales (dismissed):
This looks right to me.

craycrayfish:
> > I'm wondering if the `self.consume` should default to -1 rather than 0
> 
> @craycrayfish Don't you mean `_consume_message`? Because the docstring for _consume_message mentions that the method is supposed to be blocking. But it's not since timeout=0

I'm suggesting all the defaults should be blocking at -1 unless explicitly passed in otherwise via an argument

giddyphysicist (dismissed):
This looks good.
I would lean towards not having blocking as the default for all methods unless changed explicitly, but that's just my preference. I would prefer to have to specify wanting to block.

Barabazs:
@craycrayfish @giddyphysicist @ricrosales Let's discuss this on Slack and try to gather some more feedback.

PR #163, opened 2024-02-26
feat: decouple KafkaDataset from core logic (Runner, AbstractNode, Config,...)

Barabazs:
## Summary

This PR decouple the KafkaDataset from the core Aineko logic.

- [x] decoupling the Kafka Dataset initialisation from the AbstractNode
- [x]  decoupling the logging_dataset from the DEFAULT_KAFKA_CONFIG
- [x] decoupling the logging dataset creation from the Runner
- [x] decoupling the Kafka specific creation parameters from the Runner
- [x] generalise dataset testing in the AbstractNode
  - [x] remove the FakeKafka testing dataset
  - [x] merge the dataset test functionality in the KafkaDataset
  - [x] update the AbstractDataset to:
    - [x]  define a test flag when initialising
    - [x] add an abstract method for `setup_test_mode`
- [x] refactor the DatasetCreationStatus so it's generalised
- [x] update documentation
- [x] removed some duplicate or unused KafkaDataset attributes 

## Motivation

The core functionality of Aineko needs to be more flexible and modular before we can successfully integrate dataset plugins. 

## Author Checklist

- [x] Status checks pass.

## Reviewer Checklist


craycrayfish (approved):
LGTM, great work! Thanks for doing this, we're nearly there!

aineko/core/dataset.py
```diff
@@ -35,140 +36,115 @@ class DatasetError(Exception):
...
-            pass
-
-        def create(self, **kwargs) -> Any:
-            pass
-
-        def delete(self, **kwargs) -> Any:
-            pass
+    Subclass implementations must implement all abstract methods. Please
+    refer to the documentation of each method for more information.
+    """
```
craycrayfish, line 96: I think it would be helpful to have a summary of datasets as a concept and an  overview of all the methods that needs to be implemented and why. For example:
```
A dataset comprises 2 subcomponents, the interface layer and the storage layer. The storage layer refers to the actual storage infrastructure that holds the data, and the interface layer is an API layer that allows for the interaction with the storage layer. 

__init__: This methods controls how the dataset is initialized when instantiated
create: Creation of the actual storage layer.
read: Read interaction to query data from the storage layer.
...
```

aineko/core/dataset.py
```diff
@@ -204,3 +180,72 @@ def exists(self, *args: T, **kwargs: T) -> bool:
...
+
+        * `_input_values` for input values
+        * `_output_values` for output values
+
+        Args:
+            source_node: The name of the source node.
+            source_pipeline: The name of the source pipeline.
+            input_values: A list of input values.
+        """
+        raise NotImplementedError
```
craycrayfish, line 225: I'm thinking it might be good to set up a very simple default behavior for this so people can create a dataset quickly without having to go through too many steps. Something like setting up a list or dictionary that stores data.

aineko/core/dataset.py
Barabazs: Addressed in b6b0cd35ca81e87154496a32c83417ac55afa6e

aineko/core/dataset.py
Barabazs: Yes that could be good, but I'm not sure how feasible it is (at this point) to create a good implementation that would work for all future datasets. I would rather postpone that until we have a few more datasets, or have enough time to research and design it.

If we really want to do it now, we could do a simple implementation like the KafkaDataset with an internal list. But we should decide what to do about the message wrapper:
* do we keep it in it's current form and make a pydantic model for it?
* do we want to remove some fields?
* ...

```python
{
    "timestamp": datetime.datetime.now().strftime(
        AINEKO_CONFIG.get("MSG_TIMESTAMP_FORMAT")
    ),
    "message": message,
    "source_node": self.source_node,
    "source_pipeline": self.source_pipeline,
}
```

craycrayfish: ok i'm fine with us doing that a little later

craycrayfish (approved):
LGTM. Let's merge and release a new major version

